PR #122, opened 2026-06-08
fix env for docker CRW_AUTH__API_KEYS

mj520:
root@C20241211165165:/data/conf# docker run --rm   -e CRW_AUTH__API_KEYS=["test"]   ghcr.io/us/crw:latest
ERROR crw_server: Failed to load configuration: invalid type: string "[test123456]", expected a sequence for key `auth.api_keys`

### DIFF
--- a/crates/crw-core/src/config.rs
+++ b/crates/crw-core/src/config.rs
@@ -1102,9 +1102,42 @@ fn default_true_ext() -> bool {
     true
 }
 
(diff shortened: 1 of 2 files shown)


us:
Thanks for the fix! 🙏 Real bug, env-based CRW_AUTH__API_KEYS was definitely broken. I tidied up the commit (dropped the .gitignore bits, made it a conventional fix:) and verified it against latest main. Merging now 🚀

PR #142, opened 2026-06-13
fix(extract): stop doubling /v1 in structured-extraction chat URL

adambenhassen:
## Problem

With an OpenAI-compatible `base_url` ending in `/v1` (the OpenAI convention), the two LLM paths build the chat-completions URL differently:

| Path | Builds | Result for `base_url=…/v1` |
|---|---|---|
| Summary (`formats: ["summary"]`) | `base_url` + `/chat/completions` | `…/v1/chat/completions` ✅ |
| Structured (`formats: ["json"]` + `jsonSchema`) | `base_url` + `/v1/chat/completions` | `…/v1/v1/chat/completions` ❌ (405) |

So **no single `base_url` satisfies both** paths: dropping `/v1` fixes structured but breaks summary, and vice-versa. Gateway access logs show the doubled segment:

```
POST /v1/chat/completions      200   (summary path)
POST /v1/v1/chat/completions   405   (structured path)  ← doubled /v1
```

## Root cause

`crw_extract::structured::openai_chat_url` appended `/v1/chat/completions` to a user-supplied `base_url`, whereas the summary path (`crw_extract::llm::call_openai`) appends only `/chat/completions`. The structured path effectively assumed the `base_url` was a bare host with no version segment, contradicting both the OpenAI convention and the summary path.

## Fix

Unify `openai_chat_url` with the summary path:

```rust
match base_url {
    Some(b) if b.contains("/chat/completions") => b.to_string(), // full endpoint, verbatim
    Some(b) => format!("{}/chat/completions", b.trim_end_matches('/')), // base carries /v1
    None    => format!("{}/v1/chat/completions", default_base.trim_end_matches('/')), // bare default
}
```

- A configured `base_url` carries the API-version segment (`/v1`); only `/chat/completions` is appended.
- A base already pointing at `…/chat/completions` is used verbatim (unchanged).
- Only the bare provider **default** still gets the full `/v1/chat/completions` suffix.

Now structured extraction hits `…/v1/chat/completions`, identical to summary.

## Behavior change worth calling out

The contract for a configured `base_url` is now "include the `/v1` version segment" — same as the summary path has always required. A **bare** `base_url` (e.g. `https://api.deepseek.com` with no `/v1`) that previously worked on the structured path will now resolve to `…/chat/completions`. This is intentional: such a base was already broken on the summary path, and the goal is one base that works for both. Provider defaults are unaffected.

## Tests

- **New regression test** `openai_url_base_ending_in_v1_is_not_doubled` — pins `http://gateway:8080/v1` → `http://gateway:8080/v1/chat/completions` (fails before the fix with the doubled `…/v1/v1/…`).
- Updated the existing unit tests to the `/v1`-carrying base convention; kept the verbatim-full-endpoint and bare-default cases.
- Updated `judge_tests.rs` mock `base_url`s to a realistic `{server}/v1` so they continue to hit `/v1/chat/completions` through the shared `openai_chat_url`.

```
cargo test -p crw-extract                   # 27 + 13 + 3 + 2 pass
cargo clippy -p crw-extract --all-targets   # clean
```

## Scope

```
crates/crw-extract/src/structured.rs    | 40 +++++++++++++++++++++++++--------
crates/crw-extract/tests/judge_tests.rs |  6 ++---
```

### DIFF
--- a/crates/crw-extract/src/structured.rs
+++ b/crates/crw-extract/src/structured.rs
@@ -221,12 +221,7 @@ pub(crate) async fn call_anthropic(
     tool_name: &str,
     tool_desc: &str,
 ) -> CrwResult<(serde_json::Value, Option<LlmUsage>)> {
-    let base_url = llm
-        .base_url
-        .as_deref()
-        .unwrap_or("https://api.anthropic.com");
-
-    let url = format!("{base_url}/v1/messages");
+    let url = anthropic_messages_url(llm.base_url.as_deref(), "https://api.anthropic.com");
 
     let body = AnthropicRequest {
         model: llm.model.clone(),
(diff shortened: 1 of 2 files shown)


us:
Thanks for the sharp diagnosis and the clean fix, @adambenhassen — pinning the doubled `/v1` with a dedicated regression test is exactly the right call, and it all passes locally (`cargo test` + `clippy` + `fmt` green). 🙌

While reviewing I noticed the **Anthropic** structured path (`call_anthropic`) carries the same bug class in mirror form: it appends `/v1/messages` to a bare-host default, while the summary path (`llm::call_anthropic`) consumes `base_url` verbatim — so no single `base_url` satisfies both there either. I pushed a small follow-up commit on top of yours that unifies it with an idempotent `anthropic_messages_url()` helper (bare-host default unchanged, so zero breakage), adds unit tests, and tightens one `openai_chat_url` doc comment about the intentional `None`-branch divergence.

Your commit stays as-is — merging both now. Really appreciate the contribution! 🚀

PR #236, opened 2026-07-08
fix(renderer): CDP ws host→IP for Chromium 148+ rebinding guard (self-host chrome)

us:
## Problem
On self-host, every chrome render (including screenshots) failed with `CDP connect failed: http handshake rejected`. Verified by running the opencore docker stack locally.

## Root cause
Chromium 148+ validates the DevTools WebSocket **Host** header (DNS-rebinding guard): only `localhost` / IP literals pass. The self-host compose points the engine at the `chrome` docker **service name** (`ws://chrome:9222/`), so the WS upgrade is rejected. `--remote-allow-origins=*` (already set) only covers the Origin check, not this. The managed prod stack connects via a static IP (`172.30.40.31`), so it was unaffected — which is why this only bites self-host.

## Fix
`resolve_ws_host_to_ip()` in `cdp_conn.rs`: resolve the ws_url host to an IP before connecting, so tungstenite's derived Host header is an IP literal Chromium accepts. No-op when the host is already an IP (managed prod); best-effort passthrough on DNS/parse failure.

## Verified (local self-host, full stack)
Rebuilt the engine with the fix, ran `docker compose --profile heavy up`, scraped `fastcrw.com` with `screenshot@fullPage`:
- Before: `http handshake rejected`, no screenshot.
- After: **785×14970 full page, logo + images rendered, no raw code** — chrome connects.

Unit tests: IP is a no-op; `localhost` rewrites to a loopback IP literal. Full workspace fmt/clippy/build/test green.

### DIFF
--- a/crates/crw-renderer/src/cdp_conn.rs
+++ b/crates/crw-renderer/src/cdp_conn.rs
@@ -76,6 +76,40 @@ fn classify_ws_error(e: &tokio_tungstenite::tungstenite::Error) -> &'static str 
     }
 }
 
+/// Rewrite a `ws://host:port/...` URL so its host is an IP literal.
+///
+/// Chromium 148+ guards the DevTools WebSocket against DNS-rebinding by
+/// validating the `Host` header: only `localhost` and IP literals pass, so a
+/// connect over a docker service name (`ws://chrome:9222/...`) is rejected with
+/// an HTTP handshake error. tungstenite derives the Host header from the URL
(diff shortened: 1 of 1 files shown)


us (commented):
Verified locally: full opencore stack (engine+chrome+lightpanda), fastcrw.com screenshot@fullPage → 785×14970 full page, images render, chrome connects. Unit tests + full workspace fmt/clippy/build/test green.

PR #238, opened 2026-07-08
ci(deps): ignore ego-tree/lopdf direct bumps (parent-pinned)

us:
Stops dependabot re-opening un-mergeable direct bumps for `ego-tree` and `lopdf` — both are pinned by a parent crate (scraper 0.25 → ego-tree 0.10; pdf-inspector 0.1.3 → lopdf 0.41) whose API type flows into our code, so a direct bump can't resolve. The fix arrives via a parent-crate bump; inline comments say when to remove each ignore. Config-only.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,6 +8,17 @@ updates:
     groups:
       cargo-minor-patch:
         update-types: ["minor", "patch"]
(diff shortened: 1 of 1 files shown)


us (commented):
Config-only dependabot ignore for two parent-pinned deps; verified YAML. Merging via admin (solo repo).

PR #260, opened 2026-07-09
feat(server)!: native /v1/extract endpoint

us:
## What

Adds a **native async multi-URL structured-extraction** route. Unlike the
FC-legacy `/v2/extract` (which merges every URL's JSON into one object,
last-write-wins), the native route returns a **per-URL results array** — each
URL keeps its own object, and each carries per-URL LLM usage for downstream
billing.

- `POST /v1/extract` + `GET /v1/extract/{id}` — reuse the existing async extract
  job worker; `results: [{url, status, data, error, llmUsage}]` in request order
- Guards: `max_extract_urls` cap (default 50), per-URL SSRF preflight before any
  fetch, `basis:true` rejected pre-2b, `baseUrl` rejected with 400 (unvalidated
  SSRF vector), `llm_api_key` kept out of `Debug`
- Invalid/unsafe URLs surface as `failed` results (never silently dropped),
  preserving request order

## Surfaces

- **MCP:** `crw_extract` + `crw_check_extract_status` tools, dispatched in the
  embedded server and both proxy binaries (`crw-cli`, `crw-mcp`); shared
  `prepare_extract` helper so HTTP + MCP validate identically
- **SDKs:** TS + Python `extract()` repointed to `/v1/extract` (per-URL results);
  langchain loader emits one `Document` per URL with status/error surfaced
- **OpenAPI:** hand-authored paths + schemas (3.1 and 3.0 pairs)
- **Capabilities:** `extract.{supported,maxUrls,perFieldAttribution}`
- **Docs:** reference docs, `llms-full.txt`, and blog updated to the native route
  and the 8-tool MCP surface

## Compatibility

`/v2/extract` is externally unchanged (still the merged `data` shape) — only the
shared worker storage was extended additively.

**BREAKING (SDK):** `extract()` now returns a per-URL results array instead of a
merged object, and no longer accepts `systemPrompt` / `baseUrl`.

## Testing

- Engine: offline validation, worker-ordering, and an **end-to-end mocked-LLM**
  test (wiremock page + Anthropic tool_use); `cargo test --workspace` green
- SDKs: Python 59, TS 9
- `cargo fmt --check`, `clippy --workspace --all-targets -D warnings`, and the
  OpenAPI drift check all green

Phase 2b (per-field `basis` attribution) and the SaaS integration are separate,
gated follow-ups.


us (commented):
Self-review (solo repo — GitHub blocks self-approve). Verified locally: cargo fmt --check clean; clippy --workspace --all-targets -D warnings clean (fixed 2 pre-existing 1.97 lints in a separate chore commit); cargo test --workspace green; OpenAPI drift check green; MCP tools dispatched in all three sites; SDK + langchain + docs consistent with the native per-URL shape. Reviewed across multiple agent + external passes to consensus.

PR #297, opened 2026-07-16
fix(renderer): avoid panic slicing multibyte HTML at the scan-size cap

santhreal:
The SPA/thin-content/vendor-block detectors slice the fetched page HTML at a fixed byte cap (500KB, or 15KB for vendor blocks) before lowercasing, without walking the cap back to a UTF-8 char boundary. When a page exceeds the cap and a multibyte char straddles the cap byte, `&html[..cap]` panics with "byte index N is not a char boundary". `needs_js_rendering` runs on every non-PDF HTTP fetch (`lib.rs:1252`), so an oversized page in any non-Latin language crashes the scrape.

The same file already walks this back for the strong-marker scan, and `single.rs`/`sitemap.rs`/`structured.rs`/`antibot.rs` guard it too — these four detector sites were the only naive ones. This routes every fixed-cap prefix slice through one `boundary_capped_len` helper.

Repro: a >500KB page ending in 499_999 ASCII bytes + 'é' panics; three regression tests cover the 500KB and 15KB caps plus the helper contract.

### DIFF
--- a/crates/crw-renderer/src/detector.rs
+++ b/crates/crw-renderer/src/detector.rs
@@ -7,12 +7,24 @@ static AKAMAI_REF_RE: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"Reference #\d+\.[0-9a-f]+\.\d+\.[0-9a-f]+").expect("static regex")
 });
 
+/// Largest byte length `<= cap` that ends on a UTF-8 char boundary, so `&html[..n]`
+/// never splits a multibyte char. A fetched page longer than `cap` can place a
+/// multibyte char straddling the cap offset; slicing there would panic
+/// (`is_char_boundary` is false at that byte). Every fixed-cap prefix slice of the
+/// page HTML in this module goes through here.
+fn boundary_capped_len(html: &str, cap: usize) -> usize {
+    let mut n = cap.min(html.len());
+    while n > 0 && !html.is_char_boundary(n) {
+        n -= 1;
+    }
+    n
+}
+
 /// Heuristic: does the HTML look like an SPA shell that needs JS rendering?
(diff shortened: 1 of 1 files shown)


us:
thanks for this, and for the write-up — the repro is exactly right. `needs_js_rendering` runs on every non-PDF fetch, so this is a real crash path for any oversized non-latin page, not a theoretical one. nice catch that the strong-marker scan already walked the cap back and these four sites didn't.

reviewing it properly now (running it through the full fetch path, not just the unit boundary) and will report back shortly.

us:
merged, thanks santhreal — this was a good catch and a well-scoped fix.

i verified the panic end to end before touching it: built your repro against unpatched main and got the crash at both caps, then confirmed your fix closes it. `needs_js_rendering` runs on every non-pdf fetch, so this was a real crash on oversized non-latin pages, not a theoretical one.

two things i pushed on top of your branch:

- swapped the walk-back helper for `str::floor_char_boundary` — the stdlib has this exact function and the workspace already uses it in three places, so it sets no new rustc floor. your logic was correct, it just already ships with rust.
- reworked the tests. mutation testing showed the 512kb strong-marker site had no coverage at all (a naive re-slice there passed the whole suite), and the assertions passed even with the scan stubbed out to return nothing. the table-driven version now covers all three caps against 2- and 4-byte chars at every straddle offset, and each case asserts a verdict that fails on an empty window.

your instinct that the strong-marker scan already walked the cap back was right, and it's what pointed at the other four. also worth noting: the same bug class turned up in two more places (the extract grounding scan and the browse url logger) — both getting fixed in a follow-up that your report is what surfaced.

engine ships on releases, so this goes live with the next one.

Issue #317, opened 2026-07-18
Expose extracted URLs and images in scrape response


Antardas:
It would be useful if the scrape API returned all discovered URLs and images as structured data.

Example:

```json
{
  "markdown": "...",
  "images": [
    {
      "src": "...",
      "alt": "..."
... more metdata
    }
  ]
}
```

This would eliminate the need for consumers to parse the HTML a second time for common use cases like RAG, indexing, media downloading, and recursive crawling.

Firecrawl already has image extraction logic:
[https://github.com/firecrawl/firecrawl/blob/705bc88caee4f677681bda6fc54b7971ca1b818f/apps/api/src/scraper/scrapeURL/lib/extractImages.ts](https://github.com/firecrawl/firecrawl/blob/705bc88caee4f677681bda6fc54b7971ca1b818f/apps/api/src/scraper/scrapeURL/lib/extractImages.ts)

us:
@Antardas thank you for your feedback, will work on it quickly!

us:
hey @Antardas, thanks again for the nudge on this one. it's shipped.

scrape now takes `images` as a format and returns every image it finds as structured data, so there's no need to parse the html a second time. on the native `/v1` api you get objects with `url` + `alt` (nice as caption/embedding text for rag); the firecrawl-compatible `/v2` returns a flat list of url strings to stay a drop-in. it pulls from img src/srcset, picture sources, og/twitter/itemprop meta, icon links, video posters and inline background images.

merged in #322, going out in the next release (v0.26.0). if the shape isn't quite what you had in mind, just say the word and i'll adjust. thanks for helping make crw better.

Antardas:
Thanks @us ! This is exactly what I had in mind. Really appreciate the quick turnaround. Looking forward to trying it in v0.26.0.

PR #322, opened 2026-07-18
feat: expose extracted images in scrape response

us:
Requested in #317 by @Antardas: return discovered images as structured data so consumers don't have to re-parse the HTML for RAG, indexing, media downloading, or recursive crawling.

## What

New `images` output format:

- **`/v1`** (native): `images: [{ url, alt }]` — resolved absolute URLs plus alt text where the source carries it (handy as caption/embedding text for RAG).
- **`/v2` + `/firecrawl/v2`**: `images: string[]` — flat URL list, matching Firecrawl's `images` shape so the compat surface stays a drop-in.

Sources covered: `` src/data-src/srcset, `<picture><source>`, og/twitter/itemprop meta, icon links, `<video poster>`, and inline `background-image`. srcset is parsed per the WHATWG algorithm, so a comma inside a `data:` URI is kept whole instead of being split into phantom URLs. Deduplicated by URL in traversal order; a later duplicate upgrades an empty `alt`.

Also threaded through the CLI (`--format images`) and the MCP tool schemas.

## Notes

- Additive: `images` is only produced when requested, no change to other formats or the scrape-success path.
- The v1-objects / v2-strings split is deliberate: `/v2` stays a byte-compatible Firecrawl drop-in while native `/v1` adds the `alt` metadata.

ref #317


us (commented):
All checks green (build/test, clippy, conformance, fuzzing, CodeQL, OpenAPI + docs drift guards). Verified end-to-end on 36 live sites with a Firecrawl-parity oracle (0 divergence) plus adversarial fixtures. Merging.

PR #328, opened 2026-07-19
fix(extract): count maxChars as Unicode scalars

santhreal:
Chunk budgets are named and documented as character limits, but splitting used UTF-8 byte length. Multibyte text (CJK) was over-split under the stated cap, and sentence packing under-filled the same budget.

Count Unicode scalars for the hard cap, sentence merge, and tiny-chunk merge.

PR #329, opened 2026-07-19
fix(extract): count MAX_USER_PROMPT_CHARS as Unicode scalars

santhreal:
`MAX_USER_PROMPT_CHARS` is a 500-character cap on caller directive text, but truncation used UTF-8 byte length. A 500-character CJK directive was cut to about 166 characters under the stated limit.

Truncate by Unicode scalars for that cap in summary and answer prompt composition. Content byte ceilings are unchanged.

### DIFF
--- a/crates/crw-extract/src/answer.rs
+++ b/crates/crw-extract/src/answer.rs
@@ -7,6 +7,7 @@
 //! in the input list.
 
 use crate::llm::{self, LlmCallResult};
+use crate::summary::truncate_to_chars;
 use crate::untrusted;
 use crate::{chunking, filter};
 use crw_core::config::LlmConfig;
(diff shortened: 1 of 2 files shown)


santhreal:

santhreal:
recheck

us:
nice one santh, same clean fix on the directive cap. MAX_USER_PROMPT_CHARS was a 500-character cap that counted bytes, so a cjk directive got cut to about 166 chars, well under the stated limit. leaving the content byte ceilings untouched is exactly the right call, those are the real bill-amplification bound and bytes is the safer unit there. reviewed this alongside #328, confirmed truncate_to_chars keeps exactly the intended prefix at the boundaries (empty, at-limit, over-limit), rebased onto latest main for a fresh ci pass and will merge once checks are green. thanks for sending both of these.

us:
all checks green, merged as ea798f5. thanks for both of these, really clean work.

PR #380, opened 2026-07-29
fix(renderer): reap orphaned chrome targets on cancelled renders

us:
Chrome renderer processes accumulate on the host until an external watchdog
restarts the browser. Root cause is in the CDP target lifecycle, not the
watchdog.

## What leaks

Targets opened with `Target.createTarget` are owned by the browser, not by the
CDP client, so dropping the WebSocket does not close them. The explicit
`closeTarget` + `disposeBrowserContext` cleanup is best-effort straight-line
code, so it is skipped entirely when the fetch future is cancelled: a timeout,
a lost hedge race, a client disconnect. That is exactly when it is needed.

Prod, over one 7h window:

```
crw_target_lifecycle_total{phase="created", renderer="chrome_proxy"}  9
crw_target_lifecycle_total{phase="closed",  renderer="chrome_proxy"}  4
crw_target_lifecycle_total{phase="created", renderer="lightpanda"}   80
crw_target_lifecycle_total{phase="closed",  renderer="lightpanda"}   80
crw_chrome_pool_recycle_total{outcome="emergency_drop"}              48
```

## Change

Pass `disposeOnDetach` on `Target.createBrowserContext`. Chrome then disposes
the context itself the moment the session disconnects, closing every page in it
without running beforeunload, so the reap is unconditional rather than
best-effort. Both callers (the context pool and the legacy per-request proxy
path) now go through one params builder, with a unit test pinning the flag.

No behavior change on the success path: normal releases still close the target
and dispose the context explicitly.

## Verification

Against Chrome 150, with the context pool enabled, SIGKILLing the server
mid-render so that no Rust cleanup path can run:

```
before change   target after kill: 1   leaked
after change    target after kill: 0   clean
```

Raw CDP A/B on the same browser showed the same split: a dropped socket left the
renderer resident without the flag and reaped it within seconds with it.

`cargo fmt --check`, `cargo clippy --workspace -- -D warnings`, `cargo test
--workspace` (1503 tests) all green.

## Not covered

The legacy non-pooled, non-proxied path opens its target in the *default*
browser context, which cannot be disposed, so it is unaffected by this flag.
That path is not used in the managed stack (`config.docker.toml` enables the
pool) but is the self-host default. Closing it means giving every request its
own context, which interacts with the `cf_clearance` cookie cache, so it is left
for a separate change.

### DIFF
--- a/crates/crw-renderer/src/browser_pool.rs
+++ b/crates/crw-renderer/src/browser_pool.rs
@@ -46,7 +46,7 @@ impl ChromeConnOps for CdpConnection {
         let v = self
             .send_recv(
                 "Target.createBrowserContext",
-                json!({}),
+                crate::cdp_conn::browser_ctx_params(None),
                 None,
                 Duration::from_secs(2),
             )
--- a/crates/crw-renderer/src/cdp.rs
+++ b/crates/crw-renderer/src/cdp.rs
@@ -1794,9 +1794,7 @@ impl CdpRenderer {
                     let v = conn
                         .send_recv(
                             "Target.createBrowserContext",
-                            // No proxyBypassList: Chrome bypasses loopback by default,
-                            // which is what we want (don't route localhost via proxy).
-                            serde_json::json!({ "proxyServer": entry.chrome_proxy_server() }),
+                            crate::cdp_conn::browser_ctx_params(Some(entry.chrome_proxy_server())),
                             None,
                             Duration::from_secs(2),
                         )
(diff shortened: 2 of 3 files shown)


us (commented):
Checks: all 10 green (check, conformance, fuzzing 12m32s, CodeQL, Analyze rust + js, docs-guards, sdk-ts 22/24, cla).

Local: cargo fmt --check, clippy --workspace -D warnings, cargo test --workspace (1503 tests) all clean on the rebased branch.

Behavioral verification against Chrome 150 with the context pool enabled: SIGKILL mid-render (no Rust cleanup path can run) left a live target behind on main and none with this change. Raw CDP A/B on the same browser reproduced the same split.

PR #386, opened 2026-07-29
fix(renderer): pool the chrome_proxy tier so its targets get reaped

us:
Follow-up to #380. That change made Chrome dispose a browser context when the
CDP socket drops, which fixed the pooled `chrome` tier. It did not fix
`chrome_proxy`, because that tier never had a context to dispose.

## Why it was still leaking

`with_pool` was only ever called for the `chrome` tier, so `chrome_proxy`
always took the legacy path. That path builds a browser context only when the
request carries its own proxy. This tier's egress comes from the container's
own `--proxy-server`, so `REQUEST_PROXY` is empty and its targets land in the
*default* browser context. The default context cannot be disposed, so
`disposeOnDetach` never applies to it, and a cancelled render strands the
renderer process.

Prod, hours after #380 shipped:

```
chrome-1     (pooled)     0 stray targets
chrome-proxy (unpooled)  42 stray targets
```

chrome-guard was still firing on schedule.

## Change

Attach the same pool to `chrome_proxy`. Pool setup becomes one helper shared by
both tiers, and shutdown drains every pool instead of a single one.

Two behaviors deliberately preserved:

- **Per-request proxies are untouched.** `fetch` already bypasses the pool when
  `REQUEST_PROXY` is set and takes the legacy branch, which builds its own
  `proxyServer` context (and, since #380, disposes it on detach).
- **Country routing is untouched.** The `__cr.<cc>` exit is selected through the
  `Fetch.authRequired` credentials in `fetch_inner`, not through the browser
  context, so pooling does not change which exit a request leaves from.

## Verification

Ladder forced down to `chrome_proxy` against a real Chrome 150, then the server
SIGKILLed mid-render so no Rust cleanup path can run:

```
before change   target after kill: 1   leaked
after change    target after kill: 0   clean
```

Startup log confirms both tiers now pool:

```
browser-context pool enabled tier="chrome" pool_size=4
browser-context pool enabled tier="chrome_proxy" pool_size=4
```

`cargo fmt --check`, `cargo clippy --workspace -- -D warnings`, `cargo test
--workspace` (1504 tests) green.

## Still not covered

A CDP tier that runs with the pool disabled (`chrome_context_pool_enabled =
false`, the self-host default) still opens targets in the default context and
can strand them on cancellation. Closing that means giving every legacy request
its own context, which touches LightPanda too, so it is left for a separate
change.

### DIFF
--- a/crates/crw-renderer/src/lib.rs
+++ b/crates/crw-renderer/src/lib.rs
@@ -123,6 +123,58 @@ pub fn render_reserve(pool_size: usize) -> usize {
     }
 }
 
+/// Attach the browser-context pool to a CDP tier, when enabled and supported.
+///
+/// Every pooled tier gets its own pool, sized from the shared
+/// `[renderer.chrome_pool]` block. Beyond amortizing the handshake, the pool is
+/// what puts each render in its own browser context: the legacy path only
+/// builds a context when the request carries its own proxy, so an unpooled tier
+/// opens targets in the *default* context, which cannot be disposed and so
+/// cannot be reaped when a render is cancelled.
+///
+/// Gated off on browserless v2 per plan §"Out of scope". The backend is set
+/// explicitly in config; never URL-sniffed.
+#[cfg(feature = "cdp")]
+fn maybe_with_pool(
+    renderer: cdp::CdpRenderer,
+    config: &crw_core::config::RendererConfig,
+    tier: &'static str,
+) -> cdp::CdpRenderer {
+    if !config.chrome_context_pool_enabled {
+        return renderer;
+    }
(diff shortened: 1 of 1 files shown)


us (commented):
Checks: all 10 green (check, conformance, fuzzing 9m49s, CodeQL, Analyze rust + js, docs-guards, sdk-ts 22/24, cla).

Local: fmt --check, clippy --workspace -D warnings, cargo test --workspace (1504) clean.

Behavioral: ladder forced to chrome_proxy against Chrome 150, SIGKILL mid-render so no cleanup path runs. Stranded a target on main, none with this change. Startup log confirms both tiers pool.

PR #388, opened 2026-07-31
feat(extract): support openai-responses LLM provider

AsheTheWings:
Add transport implementation for the OpenAI Responses API protocol, supporting text generation, structured extraction, and judge calls via the /v1/responses endpoint.

Update CLI options, core configuration, and docs to accept the openai-responses provider setting.

### DIFF
--- a/config.default.toml
+++ b/config.default.toml
@@ -193,7 +193,7 @@ always_run = false
 # return HTTP 400 with a clear error pointing back here.
 #
 # [extraction.llm]
-# provider = "anthropic"        # "anthropic" | "openai" | "deepseek" | "azure" | "openai-compatible"
+# provider = "anthropic"        # "anthropic" | "openai" | "openai-responses" | "deepseek" | "azure" | "openai-compatible"
 # api_key = "sk-..."            # or use CRW_EXTRACTION__LLM__API_KEY env var
 # model = "claude-sonnet-4-20250514"
 # max_tokens = 4096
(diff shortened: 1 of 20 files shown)


AsheTheWings:

us:
@AsheTheWings thanks for that, reviewing it quickly and will merge soon

us:
thanks for this, really nice piece of work. the Responses wire format is genuinely not chat completions with a different path and you got that translation right, the flat function tools and the input_tokens/output_tokens split especially.

i reviewed it end to end and found a handful of things worth fixing before it goes in, since a merge here reaches production within the hour:

- the payload's own `status` was never read, so a `failed` or `cancelled` response arriving inside an http 200 was treated as a successful empty result
- `responses_url()` appended `/responses` after any query string, so an azure-style `?api-version=` base came out corrupted
- `call_text()` hard-errored on empty output where the sibling providers accept it, so switching `llmProvider` alone would break requests that used to work
- the http error body was echoed into the caller-facing error, which leaks the key or prompt if a gateway mirrors the request back
- both openapi.json copies and llms-full.txt still listed the old provider set

i wanted to push the fixes onto your branch but maintainer edits are off for this pr, so i opened #389 instead. it carries your commit unchanged as the first commit, with the fixes as a second one on top, and it references this pr. your authorship is intact in the history.

if you'd rather land it here, just tick "allow edits from maintainers" and i'll move the second commit over and close #389.

either way, thanks for taking the time on this one.

us:
merged as part of #389, which carried your commit through unchanged. it is on main now: ed551e3.

thanks again for building this one, the Responses transport was the right call to keep isolated from the chat completions path.

closing this in favour of #389, but the work is yours and it is in.

PR #404, opened 2026-08-05
feat(mcp): option to hide credit-billing fields from tool responses (self-hosted)

AsheTheWings:
## What

Adds `[mcp] hide_credits` (env `CRW_MCP__HIDE_CREDITS`, plus `--hide-credits` on `crw-mcp` / `crw mcp`) that recursively strips credit-billing fields — `creditCost` and `creditsUsed` — from MCP tool responses.

## Why

Credits exist for the managed SaaS billing layer. A self-hosted deployment has no consumer for them, yet every scrape/crawl/extract tool response injects them into the model's context — a few tokens per call plus attention noise. The per-response saving is small but permanent, and the operator explicitly opts in, so there is no surprise shape change for existing consumers.

## How

- **`crw-mcp-proto`**: new `strip_credit_fields()` — a recursive, shape-agnostic walk that removes the two keys at any depth (top-level `ScrapeData`, enveloped crawl page arrays, v2 `metadata` blocks). `tokensUsed` is intentionally kept: it measures real LLM provider consumption, useful telemetry for a self-hosted operator, not SaaS bookkeeping.
- **`crw-core`**: `[mcp] hide_credits` config section (default `false`), automatically exposed as `CRW_MCP__HIDE_CREDITS`.
- **`crw-server`**: `call_tool` applies the strip when configured — covers the `/mcp` endpoint **and** embedded `crw-mcp`/`crw mcp`, in both the text block and `structuredContent`. REST responses are untouched.
- **`crw-mcp` + `crw-cli`**: `--hide-credits` flag. In proxy mode the remote REST body can't be re-shaped by server config, so the strip is applied client-side after `apply_bounds`; in embedded mode the flag forces the config knob on.
- **Schemas**: `creditsUsed` moves from `required` to an optional property in the `crw_check_extract_status`/`crw_cancel_extract` MCP outputSchemas and the two OpenAPI specs (3.1 + 3.0) — otherwise a stripped body would hard-fail strict-client validation (`additionalProperties: false`). Unstripped responses still validate.
- **Docs**: `configuration.md`, `mcp.md`, `credit-costs.md`, `config.default.toml`.

## Tests

- Proto: strip covers all known key placements / no-op without credits / schema relaxed.
- Config: `[mcp]` TOML parse + `CRW_MCP__HIDE_CREDITS` env override.
- Server `/mcp` integration: with the flag on, `creditsUsed` is absent from both the text block and `structuredContent`, `tokensUsed` survives, and the stripped body still validates against the advertised schema; default config keeps emitting `creditsUsed`.
- Full `crw-server`, `crw-core`, `crw-mcp-proto` suites green; `clippy -D warnings` clean.
- Manual smoke: embedded `crw-mcp` scrape of example.com emits `creditCost: 1` by default, omits it with `--hide-credits` or `CRW_MCP__HIDE_CREDITS=true`.

## Drive-by

`Cargo.lock` on main is stale (pins workspace crates at 0.28.0 while the workspace is 0.29.0); the build re-synced it. Included here so `--locked` builds pass.

---
Maintainer edits enabled — feel free to modify/merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -778,7 +778,7 @@ checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 
 [[package]]
 name = "crw-browse"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -808,7 +808,7 @@ dependencies = [
 
 [[package]]
 name = "crw-cli"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "anyhow",
  "axum",
@@ -843,7 +843,7 @@ dependencies = [
 
 [[package]]
 name = "crw-core"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "config",
  "crw-mcp-proto",
@@ -862,7 +862,7 @@ dependencies = [
 
 [[package]]
 name = "crw-crawl"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "crw-core",
  "crw-diff",
@@ -888,7 +888,7 @@ dependencies = [
 
 [[package]]
 name = "crw-diff"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "crw-core",
  "hex",
@@ -903,7 +903,7 @@ dependencies = [
 
 [[package]]
 name = "crw-extract"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "crw-core",
  "ego-tree",
@@ -937,7 +937,7 @@ dependencies = [
 
 [[package]]
 name = "crw-mcp"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "base64",
  "clap",
@@ -953,7 +953,7 @@ dependencies = [
 
 [[package]]
 name = "crw-mcp-proto"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "jsonschema",
  "serde",
@@ -963,7 +963,7 @@ dependencies = [
 
 [[package]]
 name = "crw-renderer"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "async-trait",
  "axum",
@@ -992,7 +992,7 @@ dependencies = [
 
 [[package]]
 name = "crw-search"
-version = "0.28.0"
+version = "0.29.0"
 dependencies = [
  "crw-core",
  "futures",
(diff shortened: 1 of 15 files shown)


us:
thanks for this, and for the unusually thorough writeup. the reasoning is sound: credits are saas bookkeeping and there is no consumer for them on a self-hosted box, so paying context tokens for them on every tool call is pure waste.

we are putting it through the full review pass now (cross-surface schema check, prod blast radius, and a real end to end smoke of the embedded server). will report back with anything we find.

also confirmed your drive-by: main's Cargo.lock really is stale at 0.28.0, a local build here resyncs it exactly the way yours did. good catch.

us:
review done, and we pushed a top-up commit to your branch (81e296e). four things came out of it, all narrow:

1. the recursive strip also reached extraction output. `ScrapeData.json` and `results[].data` are filled from the caller's own jsonSchema, so someone scraping a pricing page with a `creditCost` field in their schema would have lost that value silently. the walk now skips a `json` key at any depth and `results[].data` on extract status; engine bookkeeping wrapping those payloads still goes. your test at lib.rs:1873 was locking the old behavior, so it got inverted and a preservation test added.

2. `[mcp] hide_credits` was ignored in proxy mode. `resolve_client_credentials` was already loading AppConfig and throwing everything away except `client.*`, so the config key only worked via the flag or env var. reproduced it against a local engine before and after: config-only, proxy mode, credits stayed visible. now returns the value and ORs it with the flag.

3. `--config` was not reaching the first config load, so the named file could not turn the knob off. the CRW_CONFIG export moved above `resolve_client_credentials` in both binaries, which also makes `--config` work for `client.*` in proxy mode, something that was already broken before this PR.

4. the four openapi specs are reverted. `ExtractStatusResponse.credits_used` is a plain u32 with no skip_serializing_if, so REST always emits it, and both SDKs type it required. the relaxation was only forced by a set-equality assert between the two surfaces, so that assert now checks the two directions explicitly: MCP optional, REST required. your relaxation in `extract_status_output_schema` was right and stays.

everything else held up. we checked all nine tools and yours was the only outputSchema that needed relaxing, the walk is cheap since it only touches object and array nodes, and `tokensUsed` being kept is the right call. fmt, clippy and the full suite are green at 1574.

nice piece of work, and the Cargo.lock catch was real.

us:
merged. thanks for this one, it was a genuinely good contribution: clear problem statement, opt-in by default, and you did the schema work most people would have missed entirely.

it goes out to the hosted engine on the next build, and it will be in the next release for self-hosters. `[mcp] hide_credits = true` (or `CRW_MCP__HIDE_CREDITS=true`, or `--hide-credits`) now works on the /mcp endpoint, embedded, and proxy alike.

PR #409, opened 2026-08-06
fix(extract): retry an LLM request that never reached the provider

us:
## What

A refused or dropped connection failed the whole extraction. The Responses transport sent a single POST with no retry at all; the Chat Completions path retried only HTTP 429/503 and treated every transport error as terminal.

## Why this one error and not the others

A **connect** failure is the only transport error that is safe to repeat on a POST: no connection was established, so the provider cannot have seen the body and a second attempt cannot produce a duplicate generation, or a duplicate charge. A **timeout** is excluded for exactly the opposite reason, and that contract is unchanged.

In production this is the shape a closed keep-alive connection takes: a provider or an idle load balancer drops a pooled socket, the next call picks it up, and today the extraction dies on it.

It is also what makes the Responses wire tests fail intermittently on CI. Three of the last twenty-five runs on `main` failed this way, all with `error sending request` against their own local mock server, on `07f4b96`, `01b1fc4` and `c5bd7e6`.

## Evidence

The new tests reproduce the CI failure exactly. Without the fix:

```
ExtractionError("Responses API request failed: error sending request for url (http://127.0.0.1:64930/v1/responses)")
```

which is character-for-character the shape seen on CI. With the fix both pass, and they were run 40 times without a flake.

Note the three earlier hypotheses this replaced, each disproven with a throwaway reproduction before landing anything: a stale pooled connection to a re-bound port, a client outliving the runtime that created it, and the two combined. All three are handled correctly by the client already.

## How

One shared rule (`is_undelivered`) and one shared backoff (`retry_backoff`) used by both call sites, so the two paths cannot drift apart again. Budget is unchanged at three attempts with the existing jittered backoff, and a refused connection returns faster than the 429/503 case it reuses, so the worst case still sits well inside the request deadline.

## Notes

- No API, schema or config surface change.
- `a_refused_connection_is_retried_and_then_succeeds` also asserts the server receives exactly one request, so the retry cannot be silently double-sending.

### DIFF
--- a/crates/crw-extract/src/llm.rs
+++ b/crates/crw-extract/src/llm.rs
@@ -76,6 +76,28 @@ pub(crate) fn shared_client() -> &'static reqwest::Client {
     })
 }
 
+/// Was the request never delivered?
+///
+/// A connect failure is the one transport error it is safe to repeat on a POST:
+/// no connection was established, so the provider cannot have seen the body.
+/// A timeout is deliberately excluded — there the request may well have been
+/// processed, and repeating it would double-charge the call.
+///
+/// Worth handling because a dropped keep-alive connection, an idle-timeout on a
+/// load balancer, or a provider closing a pooled socket all surface here, and
+/// failing the whole extraction on the first refused socket is a poor trade for
+/// a call that already tolerates a 30s wait.
+pub(crate) fn is_undelivered(err: &reqwest::Error) -> bool {
+    err.is_connect()
+}
+
+/// Backoff before repeating attempt `attempt` (1-based): ~0.5s, then ~1s, each
+/// with up to ~1s of jitter.
+pub(crate) fn retry_backoff(attempt: u32) -> Duration {
+    let base_ms = 500u64 * (1u64 << (attempt - 1));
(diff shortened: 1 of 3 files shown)


us:
Closing: the premise did not survive testing.

This was opened to fix the intermittent `responses_tests.rs` failure. It does not. The same test failed the same way on this branch, and the CI log settles why: the suite finished in 0.06s, while three attempts through the new backoff would have slept at least 1.5s. So the retry never fired, which means `is_connect()` was false, which means the connection WAS established and the request may have reached the server. Retrying that is exactly what the conservative predicate is there to prevent.

That also relocates the fault. The failure is the mock server going away mid-exchange, so it belongs to the test's own setup, not to the transport.

Four hypotheses were checked and discarded first, each with a throwaway reproduction: a stale pooled connection to a re-bound port, a client outliving the runtime that created it, the two combined, and a server closing a pooled keep-alive connection. The client already handles all four. The failure did not reproduce locally in 216 runs.

The transport change may still be worth making on its own terms, but not on this justification and not without a measured benefit behind it.

PR #424, opened 2026-08-09
fix: stop selling refused pages as successful scrapes

us:
Found by auditing 14 days of production logs (94,803 scrape records, 118,076 request rows, one full day of engine traces). Every finding below was reproduced live against prod before anything was changed, and again against a local build after.

## What was broken

**We returned bot walls and origin error pages as `success: true` and billed for them.** 761 requests over 14 days, across at least 6 paying accounts. Live examples at the time of the audit:

| URL | what the customer got | charged |
|---|---|---|
| `medium.com` | Cloudflare "Sorry, you have been blocked", 790 chars | 1 credit |
| `americastire.com` | CloudFront 403 country-block page, 650 chars | 1 credit |
| `docs.fastcrw.com/billing` | GitHub's "Unicorn!" 503 page, 327 chars | 1 credit |

The `>= 400` gate existed but only fired below 200 bytes, and it measured the largest of markdown/text/html/rawHtml — so an error page always cleared it, and asking for `rawHtml` made it unreachable outright. A further 470 requests were graded success with a completely empty body.

**The mirror image was also true:** a valid 68-byte `requirements.txt` came back `errorCode: "anti_bot"` with no content, because the HTTP tier decodes every non-PDF response as HTML and the file has no `<body>`. Our own docs site and `raw.githubusercontent.com` were among the top hosts hitting it.

**Crawl, batch, extract and monitor never applied the verdict at all.** A crawl of a walled site reported one completed page whose content was the block wall. The crw-saas#544 refund only ever touched the single-scrape handler, so "no billing on failure" was false everywhere else.

**A dead origin exited as three different errors** depending on which tier noticed first — 422, 500 `renderer_error` or 504. 10,107 of them were the 500.

**A transient DNS failure was served as a permanent HTTP 400** telling the caller to fix a URL that was fine.

## The threshold, and why it is not zero

A body-blind rule was the obvious fix and it is wrong. Measured on the same traffic: `stackoverflow.com/questions/tagged/rust` answers **403 with 39,499 chars of the real page**, a YouTube watch page answers 401 with 12,256, a Medium article answers 403 with 10,826. The renderer keeps those on purpose (`crw-crawl/src/single.rs`: *"the status is a soft signal, not a content gate"*), and both SDKs raise on `success:false` and discard `data`, so failing them would be a real recall regression.

The bar is 2,500 bytes because that is where the measured gap is: the largest observed error page rendered to 2,356 bytes, the next-largest 4xx body was 3,340. It is deliberately conservative and leaves large branded 404s (GitHub's is 3,340) passing. Raising it is a recall decision that needs its own benchmark.

## Benchmark impact — read before merging

On the frozen 1,000-URL run (`bench/server-runs/results-fixed-20260429T220032Z.jsonl`), 54 of the 853 successes carry `metadata.statusCode >= 400`. All 54 markdown samples were read: every one is an error or wall page, zero real content. Under the 2,500 bar **40 flip to failure and 14 stay**, so scrape-success on that artifact moves **853/1000 → 813/1000**.

That is the number going down because the old one counted walls, not because anything got worse. Truth-recall does not move: the error envelope keeps `data`, and `diagnose_3way.py` scores recall off `data.markdown` regardless of `success`. A full re-run is the pre-merge gate.

## Verified

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (1,601 passed). Against a locally built engine, real sites:

- `medium.com`, `americastire.com` → `success:false` + `errorCode:"http_error"`, body retained
- `raw.githubusercontent.com/.../requirements-server.txt` → `success:true`, 69 chars of file content
- `aave.com/docs/...` (404, 2,974 chars) → still `success:true`
- same page as `formats:["rawHtml"]` (255,904 bytes, no text format) → still `success:true`
- same page as `formats:["links"]` (no body at all) → still `success:true`
- crawl of `encuentra24.com` → `completed:2, blocked:1`, the walled page marked and cleared

## Notes for review

- `/v2` compat **semantics** change (shape does not): a Firecrawl-compat caller who received `success:true` + markdown for a small 403 now receives `success:false` + markdown, and the SaaS refunds it.
- A structural failure now reports `no_usable_content` instead of `anti_bot`. Nothing branches on `anti_bot`; refunds key on `success:false`.
- The "structured extraction requires a jsonSchema or a prompt" 400 is now skipped for a blocked URL, which answers `200 + success:false` instead.
- Deliberately **not** in this PR: clamping the JS ladder after a connect-level failure (22,741 timeouts over 14 days spend a full 30s on hosts that are dead at 2.5s, ~6.5% of the Chrome pool). It is the one change that can move recall, so it waits for its own benchmarked PR.
- The companion crw-saas PR stops billing blocked crawl/batch/monitor pages and reads the new `blocked` counter. It tolerates an engine without the field, so the two can merge independently, engine first.

### DIFF
--- a/crates/crw-cli/src/commands/crawl.rs
+++ b/crates/crw-cli/src/commands/crawl.rs
@@ -162,11 +162,17 @@ pub async fn run(mut args: CrawlArgs) -> Result<(), CmdError> {
         status: CrawlStatus::InProgress,
         total: 0,
         completed: 0,
+        blocked: 0,
         data: vec![],
         error: None,
     });
 
     // Spawn the crawl task
+    // One config load for both extraction knobs the crawl has to match the
+    // single-scrape path on.
+    let crawl_extraction_cfg = crw_core::config::AppConfig::load()
+        .unwrap_or_default()
+        .extraction;
     let crawl_opts = CrawlOptions {
         id,
         req: crawl_req,
@@ -181,10 +187,8 @@ pub async fn run(mut args: CrawlArgs) -> Result<(), CmdError> {
         jitter_factor: 0.2,
         deadline_ms_per_page: args.timeout,
         per_host_max_concurrent: 1,
-        normalize_tables: crw_core::config::AppConfig::load()
-            .unwrap_or_default()
-            .extraction
-            .normalize_tables,
+        http_retry_threshold_bytes: crawl_extraction_cfg.http_retry_threshold_bytes,
+        normalize_tables: crawl_extraction_cfg.normalize_tables,
     };
 
     let crawl_handle = tokio::spawn(async move {
--- a/crates/crw-core/src/types.rs
+++ b/crates/crw-core/src/types.rs
@@ -743,9 +743,12 @@ pub struct ScrapeData {
     /// once in `single.rs` (`FetchResult.screenshot` stays raw base64).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub screenshot: Option<String>,
-    /// Anti-bot verdict stamped once at the scrape choke (`single::scrape_url`).
-    /// `Some` means the page is a block/challenge shell — v1/v2 turn this into
-    /// `success:false`. `None` (skipped when serializing) = not blocked.
+    /// Why this document has no body: an anti-bot verdict stamped at the scrape
+    /// choke (`single::scrape_url`), or — on the crawl and batch paths, which
+    /// return documents rather than one envelope — `HTTP_ERROR_VENDOR` for an
+    /// origin error page. `Some` means the caller did not get the page they
+    /// asked for, so v1/v2 turn it into `success:false` and nobody bills it.
+    /// `None` (skipped when serializing) = a real page.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub block: Option<BlockOutcome>,
     /// The renderer snapshotted a partial DOM because the navigation budget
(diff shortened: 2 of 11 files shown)


us:
## Benchmark gate: A/B measured today, not against the frozen artifact

The frozen 1,000-URL artifact is from April, so a site that answers differently now reads as a regression against it. I built `main` and this branch as two release binaries and ran the same 200-URL sample through both at the same time, randomizing which build hits each site first (whichever goes second eats the rate limit — that biased the first attempt by one URL).

```
sample=200   before_ok=159   after_ok=142   delta=-17   (-8.5pp)
bodies present: before=180  after=181
lost (before ok -> after fail): 18
gained: 1
```

**Every one of the 18 losses is the intended cohort**: HTTP status >= 400 (403x11, 404x5, 400x1, 401x1) with a small body — the largest was 1,706 bytes. Zero unexplained flips.

Spot-checking what those 18 were returning before: `ndtv.com/world-news` 403/208 bytes, `fairwaymarket.com` 403/797, `hm.com` 403/231, `sciencedirect.com` 400/318, `welcometothejungle.com` 404/1,706. All block pages and error pages, all previously billed as successful scrapes.

**Recall is unchanged** — bodies present 180 vs 181. `diagnose_3way.py` scores truth-recall off `data.markdown` regardless of `success`, and the error envelope keeps the body, so the metric we headline does not move.

### The number to decide on

The frozen artifact predicted -4pp; measured against today's web it is **-8.5pp**, because more sites answer 4xx-with-a-wall now than in April. So a fresh 1,000-URL run should be expected to land near 79-80% scrape-success rather than 87.7%.

That is the same pages, graded honestly — the old number counted walls as successes, and Firecrawl's published figure is inflated the same way (it is `ok` on 51 of the 54 4xx rows in our frozen artifact). But it is a material change to a published marketing number and it is a founder call, not an engineering one. The honest reframe already in use ("877/921 after removing dead links") becomes the only defensible framing.

Raw data: the A/B harness and its output are kept outside the repo with the audit that produced this PR.

PR #494, opened 2026-09-03
fix(renderer): report scrape outcomes truthfully

us:
Three defects found in ten days of production request traces (2026-08-24..09-02), all of the same shape: a scrape reporting something other than what happened.

## 1. A budget-truncated render was accepted as real content

`looks_like_generic_bot_wall` passed a hardcoded `false` for the truncation flag. A body cut off mid-download has no closing `</body>`, so `body_html_without_scripts_lower` returned an empty string and the phrase list scanned nothing. The anti-bot interstitial then shipped as a successful, billed scrape.

Reproduced live against prod (engine 0.33.0) before the fix:

```
success: true            creditCost: 1
markdown: "# Security Check\n\nChecking your browser before accessing the site..."  (95 chars, entire body)
truncated: true          warning: "lightpanda_budget_truncated"
renderDecision: {"kind":"failover","chain":["lightpanda","chrome","chrome_proxy"]}
```

The flag is now a parameter, bound at each call site to the same `FetchResult` the html came from. The wall is detected and the ladder escalates instead of shipping it. `http_only` passes `false` because a body over `MAX_RESPONSE_BYTES` errors before that check, so its body is always complete.

Measured over the window: 300 customer requests received a wall body, plus 22 that were also billed as successes.

## 2. Timeout errors reported the overrun, not the budget

Ten sites built `CrwError::Timeout` from `deadline.overrun()` or `start.elapsed()` at points only reached once the budget was already spent. Both are single-digit by construction, so a caller given 30s was told `Timeout after 1ms` — 326 times in the window, every one from a caller that sent no `timeout` at all.

`Deadline` now carries `requested_ms` and those sites report it.

Sites that carry a genuine value are deliberately untouched: the per-tier `remaining` slices gated by `MIN_TIER_BUDGET`, the limiter `sleep`, and the `start.elapsed()` after a `tokio::time::timeout` really fired. `overrun()`'s doc comment now warns against the use that caused this.

## 3. The MCP tool description advertised a renderer the cloud does not run

`docs/agent-onboarding/SKILL.md` and its `mcp/crw-mcp/skills/` twin offered `"playwright"` as the `renderer` example — the one value the managed cloud rejects. Agents copied the example and got HTTP 400 (36 times in the window). It now shows `"chrome"`, and the prose docs mark playwright self-host only.

`docs/openapi.json` and the `js-rendering.md` renderer table are deliberately unchanged: the renderer stays in the enum for self-host, and both are CI-checked against it.

## Verification

- `cargo test --workspace`: 4224 passed
- `cargo test -p crw-renderer --features camoufox,cloak,cdp`: 1022 passed (the `camoufox`/`cloak` fixes are behind default-off features, so the default workspace run never compiles them)
- clippy clean in both configurations, `cargo fmt --check` clean
- Each new test was verified to fail when its fix is reverted

## Risk

The detector change makes `is_bot_wall` fire in cases it did not before, so it is the one part that touches recall. A false positive needs all three of: renderer-reported truncation, under 600 visible characters, and a phrase from the detector's list. The 600-char bail is unchanged and covered by a new test. Newly-detected walls route into the existing escalation path rather than any new code.

Worth a benchmark run against the scrape dataset before merge, since merging reaches prod within the hour.


us:
## Recall check against real traffic

Rather than reason about the false-positive risk, I measured it on four days of production traces (08-29, 08-31, 09-01, 09-02). Every truncated render in that window, classified by whether this change can touch it:

```
truncated renders            432
  no markdown                162
  over 600 visible chars     191   <- unaffected, the 600-char bail still exits first
  under 600 chars            79    <- the entire blast radius
    wall phrase matches      7     <- newly detected, previously shipped as content
    no phrase match          72    <- behaviour unchanged
```

So the change alters the outcome of **7 requests in 4 days**, and leaves 425 alone.

I then read the 72 unchanged bodies looking for a legitimate page that could be wrongly rejected. There is not one. They are login screens (`app.vellum.ai`, `acsl.au.itglue.com`), API error envelopes (`api.github.com` "Requires authentication"), a 404 (`jobs.siemens.com`), and more anti-bot walls the phrase list simply does not recognise yet:

- `pypi.org`, `jstor.org` — "Client Challenge - A required part of this site couldn't load"
- `vakifbank.com.tr` — a Turkish-language interstitial

No article, no product page, no real content in the set. The recall invariant is not at risk from this diff; if anything the measurement says the phrase list is too narrow rather than too wide, which is a separate change and not one I am making here.

Those unrecognised wall signatures are worth adding later, but each needs a corroborating signal rather than a bare phrase — the strong-marker arms ahead of `classify_block`'s length guard have no size cap, so a phrase-only arm could wipe the markdown of any page containing it.

PR #505, opened 2026-09-05
fix(batch): retain failed scrape results

### LINKED ISSUE #487: batch scrape drops a URL from the results array when its scrape errors
## What happens

When one URL in a batch scrape fails, it disappears from the job's results array entirely. The `completed` counter still advances for it, so a caller comparing `completed` against `data.length` sees a mismatch with nothing explaining it, and there is no per-URL error anywhere in the batch response to say which URL fell out or why.

The crawl surface does not behave this way. It records a placeholder for the failed page with the reason attached and counts it in `blocked`, so the caller can see exactly which URL failed and read the error.

## Where

`crates/crw-server/src/state.rs`, in the batch loop inside `start_batch_job`:

```rust
let scraped = scrape_url(...).await.ok();   // Err becomes None here
tx.send_modify(|st| {
    if let Some(mut d) = scraped {
        if d.block.is_some() { st.blocked += 1; }
        st.data.push(d);
    }
    st.completed += 1;                       // advances either way
    ...
});
```

The `.ok()` discards the error, so on the `Err` path nothing is pushed and `blocked` is not incremented, while `completed` is.

## Why it matters beyond the missing entry

`completed` is what the caller's usage is derived from, and `blocked` is what excludes a page from it. A URL that errors advances the first without the second, so it is accounted for as if it had produced a document. Getting the placeholder into the results array fixes the reporting and the accounting in the same move, because the placeholder carries `block`.

This is pre-existing and is already noted in a comment in `crates/crw-server/src/routes/v2/adapters.rs`.

## Suggested fix

The crawl path already has the shape:

- `crates/crw-crawl/src/crawl.rs` `failed_page(url, status_code, reason)` builds a `ScrapeData` with `block: Some(BlockOutcome { vendor: HTTP_ERROR_VENDOR, reason })`.
- `push_failed_page(...)` pushes it, increments `blocked`, and publishes the new state.

Both are private. Exporting `failed_page` and matching on the `scrape_url` result instead of calling `.ok()` on it would bring batch in line with crawl.

## The part that needs care

The job-completion gate keys off `completed`, so anything that changes what `completed` counts has to be checked against the gate, not just against the counter. That is why this is its own issue rather than a one-line change: it wants a test that a batch containing a failing URL still reports `completed` and still terminates.

## Reproducing

Any batch containing a URL the engine cannot turn into a document works. The simplest one today is a URL whose origin serves a body that is not a page, for example a `.docx` or an `.xls`, which now returns `unsupported_content_type` (#485). The job finishes with `data.length` one short of `completed` and no indication of which URL is missing.

atirna:
## Change

Batch scrapes now retain a blocked document when an individual scrape fails, keeping the result list aligned with the completed count while preserving the failed URL and reason.

Fixes #487

## Measured

The regression failed on current `main` with `blocked = 0` for a completed scrape error. It now reports one blocked document with the failed URL and error reason.

## Tests

- `cargo test -p crw-server start_batch_job_records_scrape_errors_as_blocked_documents -- --nocapture`

### DIFF
--- a/crates/crw-core/src/types.rs
+++ b/crates/crw-core/src/types.rs
@@ -963,6 +963,25 @@ impl ScrapeData {
         self.summary = None;
         self.chunks = None;
     }
+
+    /// Whether any page-content field survived. A document without a body is a
+    /// placeholder for a URL that produced no page (a cleared wall, or a scrape
+    /// error), as opposed to an origin error page that was kept readable.
+    /// `source_hash` and `basis` are excluded on purpose: both derive from the
+    /// fields below and never stand on their own. `screenshot` is not counted
+    /// either, so a wall captured with a screenshot still reads as no page
+    /// while the caller keeps the image.
+    pub fn has_body(&self) -> bool {
+        self.markdown.is_some()
+            || self.html.is_some()
+            || self.raw_html.is_some()
+            || self.plain_text.is_some()
+            || self.links.is_some()
+            || self.images.is_some()
+            || self.json.is_some()
+            || self.summary.is_some()
+            || self.chunks.is_some()
+    }
 }
 
 /// Typed anti-bot block verdict. `vendor` is the antibot `class_name`
@@ -1450,6 +1469,20 @@ mod tests {
         let mut d = page(404, 250);
         d.raw_html = Some("<html>".repeat(10_000));
         assert!(d.http_error().is_some());
+    }
+
+    #[test]
+    fn has_body_separates_a_placeholder_from_a_readable_error_page() {
+        let mut d = ScrapeData::default();
+        assert!(!d.has_body(), "a failed_page or cleared wall has no body");
+        d.markdown = Some("404 Not Found".into());
+        assert!(
+            d.has_body(),
+            "an origin error page kept readable has a body"
+        );
+        d.markdown = None;
+        d.links = Some(vec!["https://example.com/a".into()]);
+        assert!(d.has_body(), "a links-only format still delivered a page");
     }
 
     #[test]
--- a/crates/crw-crawl/src/crawl.rs
+++ b/crates/crw-crawl/src/crawl.rs
@@ -109,14 +109,14 @@ fn enqueue_discovered_links(
     }
 }
 
-/// Build the placeholder document a crawl returns for a URL it could not read.
+/// Build the placeholder document a crawl or batch returns for a URL it could not read.
 ///
 /// Carries only the URL, the status (0 when there was no response at all) and
 /// the reason, stamped through the same `block` field the scrape and batch
 /// paths already use, so every surface that already understands "this document
 /// is not a page you asked for" understands this one too, and the caller's
 /// `completed - blocked` billing keeps it free.
-fn failed_page(url: &str, status_code: u16, reason: String) -> ScrapeData {
+pub fn failed_page(url: &str, status_code: u16, reason: String) -> ScrapeData {
     ScrapeData {
         metadata: crw_core::types::PageMetadata {
             source_url: url.to_string(),
--- a/crates/crw-crawl/src/single.rs
+++ b/crates/crw-crawl/src/single.rs
@@ -106,6 +106,27 @@ pub async fn scrape_url(
     })
 }
 
+/// Reject the faults in a scrape template that no fetch can repair, before any
+/// network work. Shared by the single scrape and the batch route, so a bad
+/// template is one 400 on both surfaces rather than one placeholder document
+/// per URL labelled as a block.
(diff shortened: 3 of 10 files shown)


atirna:

us:
Thanks for this, and sorry for the slow start: your CI had never actually run, because a
first-time contributor's workflows sit behind an approval. That is released now.

The bug is real and I reproduced it in both directions against locally built engines rather
than trusting the unit test. Same request, two binaries:

```
POST /v1/batch/scrape {"urls":["https://example.com/","https://example.com:81/","https://example.org/"]}

main    -> completed=3  blocked=0  data=2  billable=3
this PR -> completed=3  blocked=1  data=3  billable=2
           data[2].block.reason = "Target unreachable: Could not reach https://example.com:81/: ..."
```

On main the caller receives two documents and is accounted for three pages. Your change makes
the results array and the counters agree, which is the right fix in the right place. There is
now a repro script keyed to it that exits non-zero on main and zero on this branch.

One thing worth recording for the issue: a host that does not resolve never reaches the batch
loop at all, because upfront URL validation rejects it into `invalidUrls`. The reachable cases
are the ones that pass validation and then fail during the scrape, which is what the port-81
URL above exercises.

I have pushed a commit onto your branch with three follow-ups.

**1. The failure reason did not reach the `/v2` surface at all.** `V2Document` has no `block`
field, so the retained placeholder serialized with nothing to explain it, and the route
documented to carry these was empty:

```
GET /v2/batch/scrape/{id}
  data[1] = {"metadata":{"sourceURL":"https://example.com:81/","statusCode":0,...}}   # no markdown, no warning
GET /v2/batch/scrape/{id}/errors
  {"errors":[],"robotsBlocked":[],"success":true}
```

`docs/docs/recipe-batch.md` promises "URLs that fail mid-job are recorded ... Retrieve them
after the job completes" against that route. On `/v1` your change delivers exactly that; on
`/v2` a caller was getting a silent empty document where the URL used to be absent. The
`/errors` route now emits the per-URL failures alongside the job-level one, for crawl as well
as batch, which keeps the `/v2` document shape untouched and makes the doc true.

**2. `routes/v2/adapters.rs` carried a comment describing the pre-fix accounting.** It said
`completed` "also advances for a URL whose scrape returned `Err` and pushed no document" and
called it pre-existing because "fixing it means changing what `completed` counts". After your
change an `Err` does push a document and the sum is exact, so the note is updated rather than
left to mislead the next reader into re-deriving a lower bound.

**3. The new test pinned a whole customer-facing sentence** that is defined two crates away in
`crw-crawl`, and it was the only pin on that string. It now matches on the vendor plus a
substring, the way the neighbouring test in that module does.

Two things I did **not** change, that are worth deciding on before this merges.

**The accounting change should be in the PR description.** `completed - blocked` is what the
hosted side turns into usage, so this PR changes what a batch costs. Measured on the two
binaries: a 60-URL batch where every URL fails goes from 60 billed pages to 0. That is the
intended direction and matches the "no billing on failure" rule, but it is a money path and it
should be visible in the description rather than found later.

**`err.to_string()` goes straight into a customer-visible field.** Verified live against a
build with a deliberately named sidecar:

```
data[0].block.reason =
  "Renderer error: CDP discovery failed: error sending request for url
   (http://internal-secret-chrome-host.crw-internal:9222/json/version)"
```

On a hosted deployment that renders as the internal renderer host and port. This is not
something you introduced: `/v1/scrape` returns the identical string in `error` today, and the
crawl path does the same. The cause is that `reqwest::Error`'s `Display` appends
`" for url (...)"` for every internal call, so it reaches the same field from the CDP tier, the
camoufox and cloak sidecars, and the managed LLM endpoint. `reqwest::Error::without_url()` is
the right primitive and the fix belongs at those call sites, not in the batch loop, so I am
keeping it out of this PR and handling it separately.

Smaller, not blocking: `vendor: "http_error"` for every error class means an invalid-request
rejection is labelled as an anti-bot block by `BlockOutcome::message()`, while
`CrwError::error_code()` already distinguishes `invalid_request` / `timeout` /
`extraction_error` / `shutdown`.

us:
one more commit on top. on `/v2` the retained placeholder was indistinguishable from a page, and neither sdk reads the errors route, so its block reason now rides in the document's `warning`, which is already part of the shape. the errors route lists only documents that have no body (an origin error page delivered readable is not also an error), each entry with a stable id. and a fault in the caller's own template (`actions`, or screenshot with `renderJs:false`) is now one 400 on both batch routes instead of one placeholder per url; the checks moved into `crw_crawl::single::validate_scrape_template` so the single scrape and batch share them. docs updated for `blocked` and the errors entry.

PR #527, opened 2026-09-10
feat(renderer): impersonated-http tier (Chrome TLS/JA3 via wreq)

paoloantinori:
# feat(renderer): impersonated-http tier (Chrome TLS/JA3 via wreq)

## What

A sixth renderer tier, `impersonated-http`: a plain HTTP fetcher that presents a real Chrome TLS/JA3/HTTP2 fingerprint using [wreq](https://github.com/0x676e67/wreq) (successor of rquest by the same author). No browser, no JS engine, no external service.

Two roles:

- **Auto chain hop.** Between the plain HTTP tier and the browser ladder. Fires only when the plain tier hit a wall-shaped response (anti-bot interstitial body/headers or a hard-block status: 401/403/429/503/520-530) or a fingerprint-shaped transport failure. Never on SPA-shell/thin/empty shapes (impersonation runs no JS) and never when the wall is a fingerprint-vendor challenge that needs JS (Cloudflare managed, DataDome, PerimeterX, Kasada, Imperva). The hop's accept gate is the SAME classifier the JS ladder and hedge share (`classify_js_attempt`), so a thin shell or a blocked body is a miss and the ladder runs as before. Clean sites are untouched: the plain HTTP tier stays the unconditional first fetch, so `renderedWith` and the code path for non-walled hosts are byte-identical.
- **Per-request pin.** `renderer: "impersonated-http"` (REST v1/v2 and MCP `crw_scrape`/`crw_crawl`). The one wire-level pin that does not imply `renderJs:true` (`RequestedRenderer::implies_js()`, consumed by the scrape, crawl and validation choke points), served by an early dispatch arm ahead of the render_js branch; a pinned result that fails the accept gate surfaces as an error instead of a billed success.

## Why

Sites with TLS-fingerprint (JA3) walls serve an HTTP-200 interstitial to non-browser TLS stacks. Verified live against Amazon.it (ASIN B0FHQGLXBP, 2026-09-10): the plain reqwest tier gets the "Fai clic sul pulsante qui sotto per continuare a fare acquisti" interstitial and today returns it as a billed success (`renderedWith: "http"`, 200). From the same network at the same time, curl-cffi `impersonate="chrome"` and wreq `Emulation::Chrome149` (and Chrome131) both receive the real product page. The plain-HTTP escalation triggers cannot see this shape: 200 + a small-but-nonempty body trips none of SPA/thin/empty/soft-status.

This is not an Amazon-only problem. An A/B probe from the same egress (plain reqwest vs wreq Chrome149, redirects followed, September 2026) shows the same class of first-gate TLS blocking on:

- **Temu**: plain gets a 2.9KB obfuscated anti-bot stub; Chrome149 gets the real 948KB best-sellers page.
- **Facebook public pages**: plain gets a 400 error page; Chrome149 gets the real public page (two existing open-source scrapers, `likha7/facebook-pages-scraper` and `facebook-page-info-scraper`, are built on exactly this via curl-cffi).
- Documented for Cloudflare's fingerprint tier: e.g. SillyTavern issue #5825 (OpenRouter 403s any non-browser JA3).

And equally important, the probe shows where the tier does NOT help, which the design encodes: X, Booking, Zara, LinkedIn, Glassdoor and www.reddit.com serve the same JS shell/challenge to both clients (their gate is JS execution, not TLS), and eBay's 403 after heavy probing hits both clients equally (IP rate limiting). The hop's trigger narrowing (wall-shaped only, vendor-wall suppression) exists precisely so those cases never pay for a futile impersonated request.

The pattern across vendors: TLS/JA3 is the cheapest gate, evaluated before the body; fingerprint vendors then layer a JS challenge on top. The hop covers exactly the first gate, and defers to the existing browser ladder whenever the second one is present.

## rquest is dead, long live wreq

An earlier name for this work targeted `rquest`. Every rquest version on crates.io is yanked (all 152); the project was renamed to `wreq` by the same author (0x676e67). This PR uses `wreq 0.16.1` + `wreq-util 0.2.0`, both Apache-2.0, pinned in the workspace. wreq-util presets currently top out at Chrome149 (curl-cffi's default alias is chrome150); Chrome149 and Chrome131 were both verified against the live wall.

## Shape

- `crw-renderer` cargo feature `impersonated` (default off, like camoufox/cloak): optional `wreq` (+gzip/brotli/zstd/deflate/charset) and `wreq-util`; lean builds stay dep-free and byte-identical. Forwarded from `crw-server`; the Docker image builds with `--features cdp,impersonated`; CI runs the tier tests and the server pin surface with the feature on.
- `RendererKind::ImpersonatedHttp` (serde `impersonated-http`) participates in `renderer_kind_for` and `min_deadline_for_full_ladder_ms`, but not in `js_renderers`, the breaker registry, or `has_recovery_tier` (it changes the fingerprint, not the egress IP).
- `RequestedRenderer::ImpersonatedHttp` with `implies_js() == false`; both pin-validation sites and `/v1/capabilities` use the new `available_renderer_names()` vocabulary.
- Config: `[renderer.impersonated] enabled = true` (default), `timeout_ms` (falls back to the HTTP tier timeout). The `cfg!`-folded predicate keeps the default-on flag inert in feature-less builds. No preset knob: only the Chrome family is verified against live walls; the preset is a pinned const with a live `#[ignore]` test as the sentinel for preset aging.
- The fetcher (`crates/crw-renderer/src/impersonated.rs`) shares its response tail with the plain tier (`http_only::build_http_fetch_result`: size caps, PDF sniff, binary rejection, charset-aware decode), ports the SSRF-safe redirect policy onto wreq's `redirect::Policy::custom` calling the same `crw_core::url_safety::validate_safe_url_blocking_resolved` owner (same 10-hop cap as CRW's reqwest policy), honors `REQUEST_PROXY` via a fail-closed proxy twin and the same warm per-proxy cache pattern as the HTTP tier, and DROPS fingerprint-relevant caller headers (user-agent, sec-ch-*, accept-encoding) so the header set stays coherent with the emulated JA3.
- Failover reason for the wall trigger is `VendorBlock` (`counts_for_promotion()` correctly excludes it from host learning).

## Detector scope, deliberately narrow

`looks_like_generic_bot_wall` gains exactly one phrase: the verified it-IT interstitial sentence. The English "click the button below to continue shopping" was considered and rejected: it is ordinary retail copy (empty-cart and order-confirmation pages carry it verbatim under the 600-visible-char cap) and it was never verified against a live EN wall. A false positive there would escalate to the ladder AND latch the host to proxy egress for 10 minutes. Other marketplaces' wall texts should be added only after live verification (the live `#[ignore]` probe shows how).

Note the wall is per-request intermittent: during verification the same network sometimes received the real Amazon page on plain reqwest too, and the German marketplace served its (different) robot-check page to BOTH clients that day. The live e2e test is therefore outcome-agnostic (asserts internal consistency of whichever outcome occurs); the hop fires when the wall fires.

## Tests

- Detector: the IT interstitial positive, the loose-fragment negative, the English retail-copy negative, and the real-article-quoting-the-sentence negative (600-char visible-text cap).
- Routing (`make_renderer_with_mocks`): wall -> hop -> success stamps `Failover[http, impersonated-http]`; clean site never consults the tier (panicking mock); still-blocked or thin hop result continues to the ladder; SPA shell skips the hop; pinned arm stamps `UserPinned`; pinned blocked/wall-shaped surfaces as an error; pinned request with a dead proxy fails closed.
- wiremock: clean 200 accepted, verbatim interstitial, SSRF redirect blocked, safe redirect + `final_url`, challenge-header stamping, fingerprint-header dropping.
- Serde round-trips, `implies_js` truth table (including `Auto` implying nothing), config defaults/disable/deadline math, pin-400 vocabulary (scrape and crawl, incl. the crawl `renderJs:true` contradiction), capabilities listing, MCP enum schema tests (both tools; token ceiling raised 3650 -> 3700).
- Live `#[ignore]` tests: the Amazon ASIN returns the product title via the tier; example.com + ebay.it search stay on the plain tier; an outcome-agnostic auto-chain e2e.

## Notes for reviewers and self-hosters

- MSRV: wreq needs rustc 1.98.
- If you patch the Dockerfile `CARGO_PKGS` feature list (e.g. adding `camoufox` on top of `cdp`), note this PR changed the literal to `--features cdp,impersonated`; downstream sed/replace rules matching the old literal must be updated. Prefer overriding the `CARGO_PKGS` build arg instead of editing the line.
- The btls backend builds natively on Linux with no cmake/go; cross-compile targets were not exercised here.


paoloantinori:

us:
Hey @paoloantinori thank you so much for this great pr, will review soon and merge it!

us:
thanks for the very careful writeup, this is a strong contribution. i reviewed it end to end, verified the tier live, and pushed a top-up onto your branch (then squashed the history into a single commit, still authored by you).

what i changed and why:

1. **the docker image build would have failed at release time.** `btls-sys` compiles vendored BoringSSL with CMake and generates its bindings with bindgen, so the builder needs `cmake` + `libclang`. `rust:1.98-bookworm` (buildpack-deps) ships neither, and no PR job builds the Dockerfile (`docker-build.yml` only runs on tags), so this would have surfaced first on the next release. added both to the chef stage and verified the amd64 build and the aarch64 cross-compile of btls-sys locally.
2. **the pin ran the JS ladder's accept gate.** `classify_js_attempt(..).acceptable` also fails on a body under 50 visible chars and on 404/410/500, so a pinned PDF (the pin arm runs ahead of both PDF short-circuits, `html` is empty by construction) and every 404 in a pinned crawl came back as `the response carries an anti-bot block`. reproduced against a live server on your branch: a pinned scrape of a real PDF returned that error. the pin now uses a wall-only verdict (hard-block status / bot wall / vendor / CF / challenge header, minus the structural-thinness signal), so a PDF parses, a 404 is a 404, and a thin body comes back with an actionable hint, matching what a hard browser pin does.
3. **the hop could end a chain a browser would have completed.** `acceptable` was calibrated for post-render output; the hop feeds it pre-render HTML, so any body with 50+ chars of nav/banner text was accepted. a TLS-gated SPA answers the impersonated client with its real server shell, which today reaches Chrome. the hop now also applies the auto arm's own escalation triggers (`needs_js_rendering`, thin + executable JS) to the hop body, with a routing test.
4. **budget.** the hop took the whole shared deadline with only a 500ms floor, so a tarpitting wall host (this tier's population) could leave the ladder nothing. it is now capped at half the remaining budget when a browser ladder exists, mirroring the cloak-first reserve. an HTTP-only deployment still gets the whole budget.
5. smaller ones: `Timeout` no longer counts as fingerprint-shaped; a header-announced `cf-mitigated`/`x-amzn-waf-action` challenge now suppresses the hop (its own gate rejected those anyway, so it was a guaranteed-futile round trip); `socks` added to the wreq features to match the reqwest line, since `ProxyEntry` accepts socks5 and wreq would otherwise send `Proxy-Authorization` to a SOCKS port over plain HTTP; the comment about wreq not reading env proxies was the other way round (`auto_sys_proxy` is on by default), corrected rather than disabled so a corporate-proxy self-host keeps parity with the plain tier; `impersonated-http` added to the crawl layer's low-tier list so client-rendered pages still get their second chance; the openapi enums (both copies), the `renderedWith` list, the crawl parameter table and the capabilities field contract; a `crw-cli` feature so the `--features impersonated` build command in the docs actually resolves; and CI now compiles the exact `cdp,impersonated` set the image ships, which found two stale `/v1/capabilities` assertions.

one note on your PR text: the btls backend does not build without cmake, and prod overrides `CARGO_PKGS` on its side, so the Dockerfile default governs the public image rather than the managed cloud.

merging now. thanks again, this was a genuinely well-argued change and the narrow detector scope and the fail-closed proxy twin were exactly right.

us:
shipped. merged as 8a388ed and released in [v0.35.1](https://github.com/us/crw/releases/tag/v0.35.1).

one thing worth writing down for the record, since it only showed up after the merge: the first tag's multi-arch image build failed on the arm64 leg with

```
ld: libbtls_sys.rlib(base64.cc.o): Relocations in generic ELF (EM: 62)
```

EM 62 is x86-64. btls-sys applies its own aarch64 `CMAKE_TOOLCHAIN_FILE`, but that file sets only `CMAKE_SYSTEM_NAME`/`PROCESSOR` and leaves the compiler to the environment, and because a toolchain file is present the cmake crate does not inject `CMAKE_C_COMPILER` either, so CMake used the host cc and BoringSSL came out x86-64. Fixed in #531 by exporting the per-target `CC_aarch64_unknown_linux_gnu`/`CXX_...` the cc crate reads. `ghcr.io/us/crw:0.35.1` now publishes for linux/amd64 and linux/arm64, and `GET /v1/capabilities` on the published image lists `impersonated-http`.

thanks again for the work, and for the unusually careful writeup around it.

PR #533, opened 2026-09-11
docs: TLS-fingerprint walls, where the impersonated tier helps

paoloantinori:
Follow-up to #527, documentation only.

While validating the impersonated-http tier I ran an A/B probe (plain HTTP vs the impersonated pin, same egress) across the site classes people actually ask about, and the result is worth shipping as a page rather than staying in a PR thread: the wall landscape splits into three gate shapes, and only one of them is what the tier fixes.

The new page (`docs/tls-fingerprint-walls.md`, linked from the JS-rendering pin table) covers:

- **The three gate shapes**, with verified examples and dates: TLS-only gates (amazon.it's interstitial, Temu's obfuscated stub, Facebook's 400-on-plain), JavaScript gates where impersonation cannot help (X, Booking's script challenge, Zara, LinkedIn, Glassdoor, www.reddit.com), and combined vendor gates (DataDome/Akamai) that only engage fully from datacenter egress.
- **The anti-pattern**: a block returning the identical error to two different TLS stacks is an IP-rate block (eBay demonstrated it live during the probe), and impersonation is wasted budget there.
- **A three-step diagnosis** runnable from whatever egress a deployment uses, since the landscape is egress-dependent.
- **The extension recipe** for other marketplaces: capture the wall's actual interstitial sentence, verify it is unique to the wall (the English "click the button below to continue shopping" is deliberately absent because empty-cart pages carry it verbatim), then add the exact sentence with positive and negative fixtures. This is the same discipline #527 applied to the it-IT phrase.
- **Preset maintenance**: the graceful fall-through behavior when the pinned Chrome fingerprint ages, and the `#[ignore]` live tests as the sentinel.

Local checks: `scripts/docs-guards.sh` and `scripts/check-doc-links.sh` pass (the page is registered in `docs/site.config.js`; 50 known slugs, no broken links).

### DIFF
--- a/docs/docs/js-rendering.md
+++ b/docs/docs/js-rendering.md
@@ -84,7 +84,7 @@ When `mode = "auto"` and you have multiple renderers configured (e.g., LightPand
 | `chrome_proxy` | Hard-pin to residential-proxy Chrome tier — no fallback |
 | `playwright` | Hard-pin to Playwright — no fallback |
 | `camoufox` | Hard-pin to the opt-in Camoufox stealth tier (REST) — no fallback. Requires a build with `--features camoufox` and a configured `[renderer.camoufox]` endpoint |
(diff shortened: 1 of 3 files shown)


us:
thank you for that @paoloantinori 🫶

us (dismissed):

Issue #530, opened 2026-09-10
text/plain sources: markdown format escapes backticks and collapses newlines, destroying fenced code blocks


cagerber:
## Summary

For sources served as `text/plain` (e.g. raw files on `raw.githubusercontent.com`), `formats: ["markdown"]` **escapes every backtick and drops most newlines**, destroying fenced code blocks and paragraph structure. The same URL with `formats: ["html"]` returns the body **verbatim**.

Self-hosted, default config, `ghcr.io/us/crw:latest` → `version: 0.34.1` (image digest `sha256:0e673d87b756f28c6ce432cd6dc6e3fb10547e7c18e6d59db9c7b74f6bdbbf0d`). No auth (`auth_required` unset). Reproduces on **both** `/v1/scrape` and `/v2/scrape`, with raw `curl` — no SDK involved.

## Reproduce

```bash
curl -sS -X POST http://127.0.0.1:3000/v2/scrape \
  -H 'Content-Type: application/json' \
  -d '{"url":"https://raw.githubusercontent.com/us/crw/main/README.md","formats":["markdown"]}'
```

Inspect `data.markdown`. The README's first fenced block arrives as a single line:

```
## One-command install \`\`\`bash curl -fsSL https://fastcrw.com/install | sh \`\`\` Runs local and free, no account needed. To use the Cloud, paste your key into the same command ...
```

Raw bytes after `One-command install` (`hexlify`):

```
4f6e652d636f6d6d616e6420696e7374616c6c 205c605c605c60 6261736820...
                                       ^ \` \` \`  (space, then three escaped backticks)
```

## Measured

Same URL, `data.metadata.contentType == "text/plain"`:

| variant | chars | `\n` count | backticks | backslashes | escaped backticks | ` ``` ` fences |
|---|---|---|---|---|---|---|
| direct fetch of the raw file (ground truth) | 9074 | 228 | 78 | 0 | 0 | 20 |
| `formats: ["markdown"]` | 7289 | **38** | 78 | 162 | **78 / 78** | **0** |
| `formats: ["html"]` | 9117 | 228 | 78 | **0** | **0** | **20** |
| `formats: ["rawHtml"]` | 9074 | 228 | 78 | 0 | 0 | 20 |

- **78 of 78 backticks escaped** — no fence survives.
[...]

PR #535, opened 2026-09-12
fix(extract): stop corrupting markdown for non-HTML sources

us:
## Summary

- `text/plain` (and other non-HTML) response bodies were still run through the HTML clean/readability/HTML-to-markdown pipeline. `htmd` escaped every markdown metacharacter it found and HTML's whitespace-collapse rule merged newlines into spaces, destroying fenced code blocks and paragraph structure for sources like `raw.githubusercontent.com`.
- `ExtractOptions`/`OwnedExtractInput` now carry the response content type. When it isn't HTML-like, `extract()` skips the HTML pipeline entirely and returns the body byte-for-byte for every format (`markdown`, `html`, `plainText`).
- The content-type check is shared with the renderer via a new `crw_core::is_html_like_content_type` (moved out of `crw-renderer`, which cannot depend on `crw-extract`; re-exported there for existing callers).
- Added a regression test (`text_plain_source_markdown_is_byte_for_byte_passthrough`) and verified live against the exact repro in the issue: fixed build's `/v2/scrape` output is byte-for-byte identical to the raw file (9074 chars, 228 newlines, 0 escaped backticks, 20 fences — matching ground truth), where an unpatched build reproduces the reported corruption exactly (7289 chars, 38 newlines, 78/78 escaped, 0 fences).
- Full workspace test suite (cargo test --workspace), clippy, and fmt all pass.

ref #530

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -881,7 +881,7 @@ checksum = "460fbee9c2c2f33933d720630a6a0bac33ba7053db5344fac858d4b8952d77d5"
 
 [[package]]
 name = "crw-browse"
-version = "0.34.0"
+version = "0.34.1"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -911,7 +911,7 @@ dependencies = [
 
 [[package]]
 name = "crw-cli"
-version = "0.34.0"
+version = "0.34.1"
 dependencies = [
  "anyhow",
  "assert_cmd",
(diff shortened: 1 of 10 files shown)


us (commented):
Verified live against the exact repro in issue #530: fixed build's /v2/scrape output for raw.githubusercontent.com/us/crw/main/README.md is byte-for-byte identical to the raw file (9074 chars, 228 newlines, 0 escaped backticks, 20 fences), while an unpatched build reproduces the reported corruption exactly (7289 chars, 38 newlines, 78/78 escaped, 0 fences). Full workspace test suite, clippy, and fmt all pass. All 13 CI checks green.

PR #540, opened 2026-09-14
fix(crawl): stop lowercasing the path and query of every URL crawled and mapped

gjija:
## Summary

`normalize_url` lowercased the whole URL — scheme, authority, path and query — and the crawl then **enqueued and fetched that value** rather than keeping it as a comparison key. `url_filter::normalize`, its hand-copied mirror, does the same for `/map`, where the value is also what the caller gets back.

Only the scheme and authority are case-insensitive (RFC 3986 §6.2.2.1). This folds those and leaves every other byte alone.

## Motivation

Folding the path and query produced requests nobody asked for:

- a discovered `/docs/Guide` went out as `/docs/guide`, which 404s on nginx, Apache, S3 and most app routers;
- `?id=AbC123` lost the case of an opaque token, so signed and base64-keyed URLs came back wrong or 403;
- `metadata.source_url` reported a URL that was never requested, so the caller could not see what had happened;
- `/A` and `/a` collapsed into one frontier slot, dropping one of two real pages.

`AGENTS.md` names recall a hard product invariant. This silently cost recall on every mixed-case URL served by a case-sensitive origin.

## Research

Crawlee (`apify/crawlee` @ `3a248f4c0b18230310d46c7dd9dfefadb9c4b87e`, Apache-2.0) keeps two fields on every request: `url`, byte-exact and what goes on the wire, and `uniqueKey`, normalized and used only for dedup — normalization never feeds back into what is requested. That separation is the idea this PR applies; no code was taken.

## Changes

1. **`crw-crawl`** — fold only scheme + authority in `normalize_url` and its `url_filter::normalize` mirror; `enqueue_discovered_links` now enqueues the link as the page wrote it (minus the fragment, which is never sent on the wire) instead of enqueueing its own key.
2. **`crw-server`** — `/map`'s `includePaths`/`excludePaths` fold both sides explicitly.
3. **bench** — realign the `map_url_filter` baseline, which is a hand-copy of `normalize`.

The fold is done on the string, not by re-serializing a parsed `Url`. Routing the key through the parser would *also* percent-encode non-ASCII paths, drop default ports and userinfo, apply IDNA to the host and remove dot segments — four more changes to a public `/map` response, none needed to fix the casing, all invisible in a diff that claims to be about case. It also keeps `filter_and_normalize_raw`'s sitemap fast path fast: that path pre-screens URLs with no `?` specifically to avoid a parse, on a loop sized for multi-million-URL sitemaps.

Trailing-slash handling is likewise untouched — `trim_end_matches` still runs on the whole string, so `/a/?q=1` keeps its slash. Trimming the path component instead would turn it into `/a?q=1` and relocate the does-not-resolve bug rather than remove it.

## Compatibility

`/map` responses change: discovered URLs keep the path and query case the page wrote. That is the fix. Everything else about the returned string — host form, port, userinfo, encoding, dot segments, trailing slashes — is byte-identical to today.

`includePaths`/`excludePaths` were case-insensitive **only by accident**, because discovery lowercased everything it returned. Commit 2 makes that explicit so a fix to discovery does not quietly start dropping URLs `includePaths` used to return, or leaking URLs `excludePaths` was asked to remove. Observable behaviour of both filters is unchanged.

Dedup on a case-insensitive origin (IIS, or a site whose templates spell links inconsistently) now costs two frontier slots for one document instead of one. That is the deliberate trade: the old folding dropped real pages on case-sensitive origins, which are the more common case, and the loss is bounded by `max_pages`. Called out because it is a recall trade in the opposite direction and a reviewer should weigh it rather than take my word.

## Reliability

Fixes a silent wrong-URL request. No retry, timeout or concurrency behaviour is touched.

## Performance

No measured claim, and none is made — the fold is a `find` plus a `to_lowercase` on a short prefix where it was a `to_lowercase` on the whole string. The string-based approach was chosen partly so this PR would not need a performance argument; the alternative (`Url::parse` per URL) would have needed one on the sitemap path.

## Testing

- `cargo test -p crw-crawl` — 435 unit + all integration targets pass
- `cargo test -p crw-server` — pass

Added: an integration test driving `run_crawl` against a mock origin that answers `/docs/Guide` and nothing else, **verified to fail on the parent commit**; unit tests pinning the properties this deliberately does *not* change (userinfo, default port, dot segments, percent-encoding, trailing slash before a query); the schemeless arm; and a shared corpus driven through both copies so "mirror" is a contract rather than a comment.

Three pre-existing unit tests pinned the old behaviour as contract and are rebaselined — their comments explained the lowercasing as a known quirk, which is what let it survive. The two `url_filter` tests asserting a lowercased MediaWiki `title=` are the clearest case: MediaWiki titles are case-sensitive, so preserving the case is what makes the returned URL resolve at all.

Note: four `crw-renderer` `http_only` tests fail in my sandbox (`connection_failure_catches_connect_timeout` and three blackhole/proxy tests). They fail identically on unmodified `main` here — they depend on connecting to unroutable addresses — so they are environmental, not from this change.

## Risks

- The dedup trade above.
- `/map` output bytes change for mixed-case sites. No conformance golden covers URL casing, so nothing catches this automatically either way.

## Future work

Found while doing this, deliberately left out:

- `discover_urls` still both fetches and *returns* the normalized value, so on `/map` it is a URL and not only a key. The doc comment now says that rather than claiming an invariant the code does not hold. Carrying `(key, as_written)` through it, the way `enqueue_discovered_links` now does, is the real fix.
- Relatedly, `/map` still requests `/a` for a link written `/a/`. Pre-existing; untouched here.
- `run_crawl`'s `visited` set doubles as the frontier budget, so a URL skipped by robots or a CDN error still consumes a `max_pages` slot and the crawl under-delivers against the requested limit.
- `includePaths`/`excludePaths` match as substrings of the **whole URL**, so `excludePaths: ["/api"]` matches the `//api` in `https://api.example.com/docs` and drops the host. Fixing that is a semantic change to a compat surface and wants its own PR.


gjija:
recheck

us:
Superseded by #546, which carries this change along with the rest of the series. Verified against the consolidated branch: every behaviour change and every test here is present there, and the normalizer ended up in better shape, userinfo and the port keep their case, IPv6 literals are handled, and the hand-copied `url_filter::normalize` mirror is gone in favour of a single `pub(crate) normalize_url`. Review continues in #546. Thanks for splitting the work up, it made the comparison easy.

PR #541, opened 2026-09-14
docs(research): record the competitive reading behind the current fixes

gjija:
## Summary

Adds `docs/research/competitive-analysis.md`: implementation-level notes from reading seven crawling, scraping and browser-automation projects at pinned commits.

## Motivation

Two of the fixes opened alongside this apply ideas from other projects. Without somewhere to record *which* ideas, from *what commit*, and *why the constraint applies to crw*, that reasoning lives only in a PR description and a rejected idea gets re-proposed every six months.

## Research

| Project | Commit | License |
|---|---|---|
| `apify/crawlee` | `3a248f4c0b18230310d46c7dd9dfefadb9c4b87e` | Apache-2.0 |
| `D4Vinci/Scrapling` | `48da61d1ee85cea7bbbdff013d98c90602e1d93f` | BSD-3-Clause |
| `unclecode/crawl4ai` | `862f6bccb9c063f49b9d42701baa0eea17a4993f` | Apache-2.0 |
| `firecrawl/firecrawl` | `5e57a065ac4d958399b70d11bd5fdf167bc723a3` | AGPL-3.0 (root) — `apps/api/package.json` declares ISC, contradicting it; noted in the doc |
| `ultrafunkamsterdam/nodriver` | `a71cda374651d13815a42c5eeb61af04a711eaa7` | AGPL-3.0 |
| `Kaliiiiiiiiii-Vinyzu/patchright` | `26ab9ae74516a68077f218b5de763d85be9f6d5a` | Apache-2.0 |
| `daijro/camoufox` | `571e416ac1ee52f055afc1f4e2e8ccd2b8d2ec17` | MPL-2.0 |

Every entry answers the question that actually matters — not "does this project do something clever" but **"does the constraint that made them build it also exist in crw"** — with a citation into this tree either way. READMEs were not treated as evidence.

**Nothing was copied.** crw is AGPL-3.0; each project's license and the obligations it would attach (Apache-2.0 NOTICE, BSD-3 attribution, MPL-2.0 file-level copyleft) is recorded, and none was incurred because no source was taken. Mechanisms are described in prose precisely enough to reimplement; the reimplementation is original.

Two honesty notes carried in the doc. Firecrawl's real main-content extractor is an external package **not in the public repo**, so conclusions about it are drawn from the in-tree cheerio fallback and from call sites, and are marked as such. And the browser-automation reading was scoped to robustness, process-leak avoidance and internal configuration consistency — techniques aimed at defeating an explicit access-control decision were out of scope and are not recorded.

## Changes

One new file, `docs/research/competitive-analysis.md`. Sections: repositories read; where crw is already ahead; ideas adopted; ideas adapted; ideas rejected with reasons; ranked future work; benchmark methodology.

It records what crw is **ahead** on as well as the gaps — browser context pooling, signal-robust process-group teardown, resource interception, table normalization and per-field evidence all came back with nothing to adopt, and the navigation-lifecycle axis came back empty entirely. That is a result worth keeping, not a gap to hide.

## Performance

**No benchmark numbers, deliberately.** This work produced none. The changes it accompanies are correctness fixes and none carries a performance claim. The future-work items that would move scrape success or recall are marked as needing measured before/after against the frozen corpus — which `AGENTS.md` requires before merge, not after. The doc says so explicitly rather than leaving a blank section or inventing numbers.

## Testing

`bash scripts/docs-guards.sh` — all 6 guards pass. `bash scripts/check-doc-links.sh` — 64 files scanned, 50 known slugs, no broken internal links.

## Compatibility

None. Documentation only, no code.

The file lives outside `docs/docs/` deliberately: it is engineering notes, not a page of the docs site, so it is not in `site.config.js` and generates no HTML. Per `docs/AGENTS.md` that also means there is nothing to regenerate. Say the word if you'd rather it live elsewhere (`docs/adr/`, or a root `RESEARCH.md`) and I'll move it.

## Risks

It dates. The ranked future-work list is the part most likely to go stale, and the commit SHAs pin what was read but not what those projects do now.

## Future work

The doc's own ranked list, briefly: deterministic CSS/XPath schema extraction (removes the LLM key from the structured-extraction critical path — both engines are already in-tree); per-node content pruning as one more rung in the existing quality ladder, benchmark-gated; concurrent crawl dispatch (`run_crawl` builds a semaphore and then awaits each fetch inline, so real concurrency is 1 and `max_concurrency` is inert — the identical bug was already found and fixed in the `/map` path, with the reasoning recorded in-file); an overall crawl deadline; proxy health scoring; adaptive per-host pacing.


gjija:
recheck

us:
Closing this one for a different reason than the rest of the series. The reading behind these fixes is genuinely useful, but it is already captured where it gets read: in each PR description and in the code comments at the call sites that act on it. A standalone 300-line research document in the tree goes stale the first time one of those seven upstreams refactors, and nothing in CI can tell us when that happens. The content stays available on this PR if we want to point at it later. Nothing against the work, we just do not want it as a tracked file. Thanks for writing it up.

PR #542, opened 2026-09-14
fix(crawl): honour robots rules keyed on the query, and fix two parser bugs

gjija:
## Summary

Three robots.txt defects, all in the **fail-open** direction — each one let the crawl fetch something the site had forbidden.

## Motivation

**1. `run_crawl` matched on the path only.** It called `robots.is_allowed(parsed.path())`, dropping the query. Rules keyed on a query string therefore matched nothing. Hacker News disallows `/hide?`, `/vote?` and `/reply?`; `/hide?id=1` was fetched anyway.

The query-aware `is_url_allowed` already exists, its doc comment describes this exact failure using that exact example, and `discover_urls` already uses it. The crawl — the surface that actually fetches at volume — was never updated.

**2. Consecutive `User-agent:` lines lost all but the last.** RFC 9309 §2.2.1: a run of agent lines heads **one** group, and the group applies if any of them matches. Reading the flag from the last line alone discarded the rules of a very common shape:

```
User-agent: *
User-agent: BadBot
Disallow: /
```

The last agent is not us, so a site-wide `Disallow` addressed to us via `*` was thrown away.

**3. `directive_value` sliced the original line at an offset validated against a lowercased copy.** `to_lowercase` is full Unicode and does not preserve byte length — `İ` U+0130 is 2 bytes and lowercases to 3, `K` U+212A is 3 bytes and lowercases to 1 — so the offset could land past the end of the line or mid-character and panic on remote attacker-controlled input.

To be precise about severity: **no live panic was reachable.** None of the four prefixes in use (`user-agent:`, `sitemap:`, `disallow:`, `allow:`) contains a letter with a length-changing uppercase mapping. It became a remote panic the day someone added a prefix containing `k`.

This is politeness and legal exposure, not a performance nit.

## Changes

- `crawl.rs` — one line: `is_allowed(parsed.path())` → `is_url_allowed(&parsed)`.
- `robots.rs` — track whether the previous directive was also a `User-agent:`, so a run of them forms one group. `Sitemap:` is a non-group record (§2.2.3) and may appear between agent lines, so it deliberately does **not** close the run.
- `robots.rs` — compare the directive prefix in place with `eq_ignore_ascii_case`, which is byte-length-preserving by construction; `get` returns `None` instead of panicking on a mid-character split. The per-line allocation goes away as a side effect.

## Reliability

All three failure modes were "we fetched a URL the site told us not to". Nothing here can cause a page to be *missed* that was previously fetched, except where the site explicitly forbade it.

## Performance

No claim. One allocation per `robots.txt` line disappears; `robots.txt` is fetched once per crawl, so this is not worth measuring and none was done.

## Testing

- `cargo test -p crw-crawl` — 428 unit + all integration targets pass

The integration test counts requests to the disallowed path with `.expect(0)`, so a crawl that ignores the rule fails on the **count** rather than on an absence. **Verified to fail on the parent commit** ("a page disallowed by a query-keyed robots rule was crawled").

Unit tests cover: both group shapes (`*` first and our name last); that a rule line closes an agent run while a sitemap line does not; that a fresh group does not inherit the previous group's match; the length-changing Unicode characters that would have tripped the old slicing; and that directive prefixes stay case-insensitive after the rewrite.

The pre-existing `robots_multiple_user_agents` integration test, which pins that a `Googlebot`-only group must not bind us, stays green — blank-line-separated groups behave as before.

## Compatibility

Behaviour change on purpose: crawls of sites using query-keyed rules or multi-agent group headers will now skip URLs they previously fetched. No API or config surface changes.

## Risks

The group-run change is the one to look at hardest. A `robots.txt` whose author intended two separate groups but omitted a rule line between the agent headers now sees them merged — which is what RFC 9309 mandates, but it is a behaviour change on parsing input we do not control.

## Future work

Deliberately not changed, because both alter *which* rules apply rather than whether they are read at all, and each wants its own decision:

- Agent matching is `contains("crw")`, so a group naming a different bot whose token contains that substring would bind us.
- Rules from every matching group are merged into one list and compete on pattern length, rather than the single most specific group being selected as §2.2.1 requires.
- `RobotsTxt::fetch` returns an empty (allow-everything) ruleset for **any** non-2xx, including 5xx and transport errors. RFC 9309 §2.3.1.3 says an unreachable robots.txt should mean the site is fully disallowed, and the rest of this codebase fails closed. Changing it would make crawls of sites with flaky robots.txt start failing, so it needs its own PR and a deliberate call.
- `Crawl-delay` is never parsed, and `per_host_min_interval_ms` is a documented config field that nothing reads.


gjija:
recheck

us:
Superseded by #546. Verified against the consolidated branch: the query-aware robots check, the consecutive `User-agent:` grouping, the `Sitemap:` non-group record handling and the `directive_value` byte-length fix are all there, together with all seven tests from this PR. #546 goes further and replaces the substring `contains("crw")` agent match with exact product-token precedence. Review continues in #546.

PR #543, opened 2026-09-14
fix(renderer): remove self-deadlock and panic paths in the browser context pool

gjija:
## Summary

Three concurrency-correctness defects in `BrowserContextPool`
(`crates/crw-renderer/src/browser_pool.rs`), all on paths that run during
cancellation or shutdown. Three separate commits, so any one can be reverted
without the others.

1. `acquire()` could **self-deadlock** on a stale non-`Idle` slot.
2. `dec_inflight_and_notify` used a **non-atomic check-then-act**, so it could
   produce exactly the `usize::MAX` wraparound its own comment says it prevents.
3. `PoolGuard::drop` **unwrapped a poisoned mutex**, violating the panic-free
   invariant the module documents for that path.

No happy-path behaviour change, no public API change.

### 1. Self-deadlock in `acquire()` — the significant one

```rust
let conn = match &s.lock().unwrap().state {
    SlotState::Idle { conn, .. } => conn.clone(),
    _ => {
        self.mark_slot_dead_and_drop(&s);   // re-locks the same StdMutex
        continue;
    }
};
```

A temporary in a `match` scrutinee lives until the end of the whole `match` —
edition 2024's rescoping (RFC 3606) covers `if let`, not `match` — so the `_`
arm still held the `MutexGuard` when it called `mark_slot_dead_and_drop`, which
locks the same non-reentrant `StdMutex`.

That blocks an **OS thread**, not a future: it parks a tokio worker permanently
and strands the `OwnedSemaphorePermit` the caller already holds, so the pool
loses a slot for the life of the process.

**The `_` arm is reachable, not dead code.** `acquire()` pops the slot from the
idle deque (releasing `idle`); `shutdown()` phase 3 reaches the same slot
through `all_slots` and `take_conn()`s it to `Closing`. `inflight` is not
incremented until after this block, so the drain loop can observe zero and enter
phase 3 while an acquire is mid-flight — no nanosecond race required. And
`needs_check` requires `last_used > 60s`, which is the normal state of a warm
pool at SIGTERM. The identical race is already handled gracefully 40 lines
later, which is evidence the original author expected it here too.

### 2. Non-atomic saturating decrement

```rust
let prev = self.inflight.load(Ordering::SeqCst);
if prev > 0 { self.inflight.fetch_sub(1, Ordering::SeqCst); }
```

The comment above it says this prevents "a wraparound to `usize::MAX` that would
make the shutdown drain loop spin forever". It does not — two callers observing
`prev == 1` both subtract. This code exists *only* as defence-in-depth against
that case, so it should actually work.

### 3. Panic in `Drop`

`BookkeepingToken` documents: *"Drop runs during future cancellation in tokio.
MUST NOT panic — no unwrap"*. `PoolGuard::drop` then unwrapped the slot mutex
twice and reached `refresh_gauges` (`idle.lock().unwrap()`). A panic in `Drop`
while unwinding aborts the process.

The protection was on the wrong function — `record_target`, not a destructor,
was already poison-tolerant. The codebase knew: the pre-existing test carried
`slot.clear_poison(); // so the guard's own (non-poison-tolerant) Drop doesn't
panic`. That workaround is deleted here, and the test passing without it is
extra evidence.

## Changes

- `acquire()`: snapshot the connection in its own block, match on the owned
  `Option<Arc<C>>`, so no guard is live across `mark_slot_dead_and_drop`.
  Control flow is otherwise identical — `continue` still targets and advances
  `for attempt in 0..3`, and `permit` is a loop-outer binding neither arm moves.
- `dec_inflight_and_notify`: single `fetch_update` RMW with `saturating_sub`.
  `SeqCst` on both orderings matches every other `inflight` access.
- `PoolGuard::drop` + `refresh_gauges`: `unwrap_or_else(|p| p.into_inner())`.

**Scope deliberately excludes `release()`.** An earlier revision of this branch
also converted four lock sites there; that was removed. `release()` is an
`async fn` where a poisoned slot is an ordinary task panic, not an abort, and
silencing it routes the slot into the `other =>` arm that records
`record_recycle_outcome("shutdown_raced")` — mislabelling it in production
metrics and turning a loud failure into a silent CDP connection leak.

## Reliability

Fixes a permanent worker-thread park plus permit leak, a shutdown-drain spin,
and a process abort — all on cancellation/shutdown paths.

## Performance

No measurable change; the same atomic and lock operations, re-shaped. **No
benchmark claim is made.**

## Testing

Three regressions, each checked against the unfixed code before being kept:

| test | vs. unfixed code |
|---|---|
| `acquire_does_not_deadlock_on_stale_non_idle_slot` | **FAILED** after 10s |
| `guard_drop_survives_a_poisoned_slot_mutex` | **FAILED** |
| `over_decrement_floors_at_zero` | passed — see note |

The deadlock test is a plain `#[test]` driving `acquire()` on a dedicated thread
with an mpsc watchdog, **not** `#[tokio::test]` + `tokio::time::timeout`: the
bug blocks the OS thread, so a timeout future on that runtime can never be
polled to observe it, and the first version of this test simply hung. The 10s is
a failure-path ceiling only — the green path returns in microseconds. It
distinguishes `Timeout` from `Disconnected` so an unrelated panic is not
reported as a deadlock.

**Stated plainly:** `over_decrement_floors_at_zero` does **not** reproduce the
race it fixes — the old window is too narrow to hit from a unit test and it
passes against the unfixed code. It pins the saturation contract and does
deterministically catch a regression to a bare `fetch_sub(1, ..)` (1 − 16 wraps
every time). The guarantee against the race comes from `fetch_update` being one
atomic RMW, not from that test.

```
cargo test -p crw-renderer --features cdp --lib browser_pool
test result: ok. 51 passed; 0 failed

cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
bash scripts/check-no-process-exit.sh
bash scripts/check-internal-dep-versions.sh
python3 scripts/release/audit_release_please_config.py
python3 scripts/release/test_guards.py      # 14/14
```

**Not run, and I can't run it here:** the scrape-success / recall benchmark gate
that `AGENTS.md` requires before merging renderer changes — it needs live
network and a real Chrome, neither available in my environment. I believe the
risk is one-directional (un-deadlocking `acquire()` can only increase successful
renders; the `inflight` change only affects shutdown drain latency), but that is
an inference, not an attestation, and the gate is yours to run.

Also note: `crw-renderer`'s four `http_only` connect-timeout tests
(`connection_failure_catches_connect_timeout` and three siblings) fail in my
sandbox **on unmodified `main` as well** — an environment artifact of the
outbound proxy, unrelated to this branch.

## Compatibility

None affected. Private internals of `crw-renderer`.

## Risks

Low. The `acquire()` change is a pure scope change with identical control flow.
`Drop` now proceeds on a poisoned slot where it previously aborted — deliberate,
and the same choice `record_target` already made.

## Future work

Found while reading this module, deliberately not here:

- `idle_timeout` and `recycle_after_navs` are plumbed config → `PoolCfg` and
  never read. There is no idle reaper at all, and `crw-core`'s metrics document
  an `idle_timeout` outcome label that is never emitted. Implement or delete.
- `try_hedge`'s loser drops into `PoolGuard::drop`, which tears the CDP
  connection down rather than recycling the context — so every hedge LightPanda
  wins destroys a pooled Chrome connection. Benchmark-gated per the crate's
  `AGENTS.md`.
- `ClearanceCache`'s `entries`/`locks` maps are unbounded; `preference`,
  `egress` and `breaker` all use `moka` with capacity + TTL.
- Seven poison-tolerant lock sites now use three different idioms; a
  `lock_slot()` helper would put the "why" in one place.


gjija:
recheck

us:
Superseded by #546. Verified against the consolidated branch: the `fetch_update` single-RMW decrement, the health-check snapshot scope fix, the poison-tolerant `Drop` and all three tests from this PR are present there. Review continues in #546.

PR #544, opened 2026-09-14
fix(extract): stop deleting § and ¶ from body prose during markdown conversion

gjija:
## Summary

`strip_anchor_artifacts` deleted every `¶` and every ` §` from the **entire**
markdown document, not just from header anchor links. That silently destroyed
section signs in body prose — `See § 230 of the CDA` came out as
`See 230 of the CDA`.

The fix is the deletion of those two `replace` calls. The anchor regex is
unchanged.

## Motivation

`crates/crw-extract/src/markdown.rs`:

```rust
let cleaned = EMPTY_ANCHOR_RE.replace_all(md, "");
cleaned
    .replace('\u{00b6}', "")   // pilcrow ¶
    .replace(" \u{00a7}", "")  // section sign § (preceded by space)
```

`EMPTY_ANCHOR_RE` (`\[¶?\]\(#[^)]*\)`) already removed the anchor form, so the
two `replace` calls only ever reached body text, table cells and code fences.

Measured on `main`, through `html_to_markdown`:

| input | output on `main` |
|---|---|
| `<p>See § 230 of the CDA for details.</p>` | `See 230 of the CDA for details.` |
| `<p>Nach § 1 BGB gilt das.</p>` | `Nach 1 BGB gilt das.` |
| `<p>The pilcrow ¶ marks a paragraph break.</p>` | `The pilcrow  marks a paragraph break.` |

The space is consumed too, so nothing even marks where the character was.
Section signs are load-bearing in statutory content (US code citations, German
and EU law), so this is content loss, not cleanup — and `AGENTS.md` makes recall
a hard product invariant.

## Changes

One change: delete the two document-wide `replace` calls. The doc comment is
rewritten to say which artifact is removed and why the scope is the anchor link
rather than the document.

**Considered and rejected:** also widening the regex to `\[[¶§]?\]\(#[^)]*\)` so
it strips `[§](#id)`. That would reintroduce the same class of content loss this
PR removes — a lone `§` linked to a fragment is the standard statutory
cross-reference shape on LII, Justia and gesetze-im-internet.de, and
`See [§](#sec230) for details.` would become `See  for details.` A PR whose
argument is "section signs are load-bearing in legal content" should not ship a
rule that deletes legal cross-references. Left as-is.

## Reliability

Direct recall fix: statutory citations, and any prose discussing `¶`/`§`, now
survive extraction intact.

## Performance

Marginally less work — two fewer whole-document scans and allocations per
conversion, and the function returns the regex's `Cow` via `into_owned()`
instead of chaining two more `String` allocations. **Not measured; no
performance claim is made.** This PR is about correctness.

## Testing

Three new tests in `markdown::tests`, **all verified to fail against `main`**:

- `preserves_section_sign_in_body_prose` — `§ 230` and `§ 1 BGB` survive.
- `preserves_pilcrow_in_body_prose` — `pilcrow ¶ marks` survives.
- `preserves_a_decorative_sigil_that_is_not_an_anchor_link` — pins the
  trade-off below, so the fix is not undone by reinstating a global replace.

The existing `strips_pilcrow_from_headers` and `strips_empty_anchor_links` are
untouched and pass before and after: the `¶` headerlink case was always the
regex's work, never the global replace's.

```
cargo test -p crw-extract
test result: ok. 652 passed; 0 failed; 2 ignored
(+ 16 integration binaries, all green)

cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
bash scripts/check-no-process-exit.sh
bash scripts/check-internal-dep-versions.sh
python3 scripts/release/audit_release_please_config.py
python3 scripts/release/test_guards.py      # 14/14
```

I also checked that nothing depends on the old stripping: no `¶`/`§` (or
`&para;`/`&sect;`/numeric-entity/percent-encoded forms) appears in any fixture
tree — `crates/*/tests/fixtures/`, root `tests/fixtures/`, `bench/fixtures/`,
`conformance/fixtures/firecrawl_v2/` — and the conformance job compares response
**shape** (keys + value types), not markdown text, so it cannot move.

## Compatibility

Behavioural change to markdown output, deliberately: a `¶` or `§` that is not
part of a fragment anchor link is now preserved.

One case is genuinely narrowed: a bare sigil a site renders as header decoration
*without* an anchor (`<h2>Section <span class="anchor">¶</span></h2>`) now
appears in the output. The two cases are indistinguishable after conversion, so
the trade is resolved in favour of not destroying content, per the recall
invariant. Slightly wider than that: a permalink whose `href` is not a bare
fragment (`[¶](/docs/page#anchor)`) also keeps its sigil now.

No API, config or wire-format change.

## Risks

Low, and asymmetric: the risk in one direction is a stray decorative sigil on
some doc sites; in the other it was destroyed legal citations on every page that
cites a statute.

A narrower variant — strip only a trailing sigil at the end of a heading line —
was considered and not taken: it adds a line-shape heuristic that mis-fires on
`## § 230 Immunity`, which is a real heading.

## Future work

Found in this file and nearby, deliberately not here:

- `strip_data_uris`'s regex is case-sensitive and disallows leading whitespace,
  so ``'&lt;img src="DATA:image/png;base64,..."&gt;'`` and `src=" data:...">` still put
  multi-MB blobs in the markdown. `clean.rs` has the same two gaps;
  `readability::extract_images` does it correctly, so the paths disagree.
- `convert_indented_code_to_fenced` toggles `in_fenced` on any ``` fence, but
  htmd sizes a fence as `3.max(longest_run + 1)`, so a code sample containing a
  fence line desynchronises the state.
- A multi-block `<li>` (two `<p>`s) is indented 4 spaces by htmd and gets
  re-fenced as a code block, demoting prose to code.
- `extract_links` filters `javascript:` case-sensitively while `extract_images`
  lowercases first, so `href="JavaScript:alert(1)"` reaches the `links` output.
- `strip_data_uris` still does `.replace_all(..).to_string()` where
  `.into_owned()` would avoid a copy.


gjija:
recheck

us:
Superseded by #546. Verified against the consolidated branch: the document-wide `§` and `¶` replaces are gone and all three regression tests from this PR are present. #546 scopes the pilcrow strip more narrowly still, to a linked pilcrow at the end of a heading line, so a `[¶](#id)` inside body prose now survives too. Review continues in #546.

PR #545, opened 2026-09-14
fix(renderer): bound the response body read instead of measuring it afterwards

gjija:
## Summary

Both HTTP tiers checked `Content-Length` against `MAX_RESPONSE_BYTES` and then read the whole body in one call. The size was re-checked in `build_http_fetch_result`, so an oversized response produced the right *error* — but only after every byte was already resident in memory.

## Motivation

`Content-Length` cannot bound this, because in the two cases that matter **the header is simply absent**:

- **Chunked responses.** `Transfer-Encoding: chunked` is the default for any streamed or dynamically generated body, and `content_length()` is then `None`.
- **Transport-decompressed responses.** These clients enable gzip/brotli/deflate, and tower-http's decompression body reports `SizeHint::default()` — so `content_length()` is `None` there too.

That second point is worth being precise about, because my first draft of this PR got it wrong and said `content_length()` returns the *compressed* size. It does not. There is **no number at all**. A 1 MiB payload that expands to gigabytes arrives with nothing to check it against. Both new tests now assert `content_length() == None` so the premise cannot silently stop holding.

Either way a single origin — or a redirect to one — could make the process allocate until it died, taking every concurrent request with it.

## Changes

- `http_only::read_body_capped(resp, max)` — streams and checks as it accumulates. Peak is bounded at roughly one chunk over the cap, and the cap is measured **after** decompression because it applies to the bytes the reader yields. Dropping the response on the error path leaves the body unread, so the connection is not returned to the pool and the origin stops sending.
- The impersonated tier has the identical bug and gets the identical fix.
- The cap is a **parameter**, not a constant, so the boundary is directly testable. Both production call sites pass `MAX_RESPONSE_BYTES`.

The buffer is deliberately **not** pre-sized from `Content-Length`: sizing from a header lets any origin make us allocate the whole cap by advertising it and then sending one byte.

## Reliability

I checked `chunk()` against `bytes()` in the vendored reqwest source: both drive the same body, both map errors through `crate::error::decode`, and `reqwest_message` strips the URL that `do_bytes` additionally attaches. Messages and error classification are byte-identical, and there is no input where one succeeds and the other fails. **No recall change** — no page fetched today stops being fetched.

## Performance

No measured claim. Accumulating into a `Vec` copies each chunk where `Collected::to_bytes()` could return a single-frame body zero-copy, so large bodies pay roughly one extra pass of memcpy. That is the deliberate cost of not buffering unboundedly first, and it is not on any hot path worth a benchmark.

## Testing

- `cargo test -p crw-renderer --features cdp,camoufox,cloak,impersonated` — 1029 pass
- `cargo clippy -p crw-renderer --all-targets --features cdp,camoufox,cloak,impersonated -- -D warnings` — clean

Tests drive `read_body_capped` directly and run in under a tenth of a second: a chunked body over the cap; the **inclusive boundary** at exactly the cap and one byte past it, pinned against `build_http_fetch_result`'s own `len > MAX` so the two cannot drift; a **gzip payload under the cap on the wire and over it decompressed**; reassembly across chunks; an empty body.

Two notes on how these got there. The obvious assertion — that the fetch returns `Response too large` — is **vacuous**: the post-hoc check already produced that error with or without the cap, so only the memory was ever wrong. An earlier draft asserted on bytes-consumed from a 150 MiB loopback transfer, which worked but is not something that belongs in a unit test that runs on every `cargo test`. Parameterizing the cap made both problems go away.

Note: four `crw-renderer` `http_only` tests fail in my sandbox (`connection_failure_catches_connect_timeout` and three blackhole/proxy tests). They fail identically on unmodified `main` here — they depend on connecting to unroutable addresses — so they are environmental, not from this change.

## Compatibility

No API or config change. The error message for an oversized body becomes `Response too large: exceeds N bytes` (previously `Response too large: N bytes (max M)`). The escalation gate in `lib.rs` matches on `starts_with("Response too large")`, so the ladder is unaffected — but it is a user-visible string change and worth knowing.

The `wreq` lockfile delta is exactly the two edges `stream` adds (`sync_wrapper`, `tokio-util`). Unrelated: the committed `Cargo.lock` carries the workspace crates at `0.34.1` while `Cargo.toml` says `0.35.1`, so cargo rewrites those eleven lines on every invocation and `--locked` fails on `main` today. I reverted that churn out of this commit since versions belong to release automation, but it's worth a separate look.

## Risks

- `impersonated` is off by default, so its new loop is exercised only by the feature-enabled CI job. I ran clippy with the feature set explicitly because no CI job does.
- The `stream` feature on wreq pulls `tokio-util` and `sync_wrapper`. Neither touches TLS or the emulation profile, and the lean build never compiles wreq at all, so the byte-identical-lean-build claim in the manifest still holds.

## Future work

- **This is now the third near-identical helper in the workspace.** `crw_crawl::sitemap` has one under this exact name and `crw_search::client` has `read_capped`. Hoisting one copy into `crw-core` — which already owns `reqwest_message` — is the right follow-up; I left it out to keep this reviewable.
- **Sibling body reads with no cap at all**, deliberately out of scope here but worse than what this fixes: `crw-crawl/src/robots.rs` does a bare `resp.text()` on a fully attacker-controlled origin, for every host in a crawl, with no header check either — and RFC 9309 §2.5 specifies a 500 KiB limit. `crw-renderer`'s `camoufox.rs` and `cloak.rs` read page bodies with no cap, so after this PR 2 of the renderer's body-producing tiers have one and the rest do not.
- `build_http_fetch_result`'s own size check is now unreachable from both callers; keeping it as defence-in-depth is defensible but should be a deliberate call.
- For PDFs the caller now owns a `Vec<u8>` and `raw_bytes` is `Option<Vec<u8>>`, so `bytes.to_vec()` could become a move — a free ~50 MiB saved per large PDF.


gjija:
recheck

us:
Superseded by #546. Verified against the consolidated branch: the streaming size check on both transports, the `stream` feature on wreq and the chunked, boundary, decompressed and reassembly tests are all present. One small thing did not make the jump, `read_body_capped_accepts_an_empty_body`, which is worth re-adding in #546. Review continues there.

PR #546, opened 2026-09-14
fix: harden crawl and renderer reliability

gjija:
## Summary

This consolidates the validated reliability fixes from the contributor's related pull requests into one reviewable change set.

- Preserve case-sensitive URL paths, queries, userinfo, and ports while normalizing only schemes and hosts.
- Apply robots.txt rules to paths plus queries, follow exact product-token precedence, and cap decoded robots data at 500 KiB.
- Treat unavailable robots files separately from unreachable origins, with fail-closed behavior when policy enforcement is enabled.
- Prevent browser-pool deadlocks, counter underflow, and cleanup panics.
- Preserve legal section signs and body pilcrows during Markdown conversion.
- Bound decoded response bodies for both HTTP transports before oversized content is buffered.

## Compatibility

The public API shape is unchanged. Path and query casing now remain intact, robots policy failures become explicit terminal failures when enforcement is enabled, and oversized decoded bodies return the existing response-size error.

## Testing

The required commit hook completed successfully:

- cargo fmt check
- cargo clippy with warnings denied
- cargo build
- cargo test for the full workspace

The suite includes local end-to-end fixtures for crawl robots enforcement, discovery timeout behavior, Markdown extraction, chunked bodies, and decoded gzip limits.

## Attribution

The consolidated contributor commit retains `gjija <agonavdimetaj@gmail.com>` as its author. Maintainer hardening remains in a separate commit.

Refs #540, #542, #543, #544, #545.


gjija:

us:
recheck

us:
Heads up: I rewrote the branch, so please `git fetch` and `git reset --hard` your local copy before your next push.

Two things happened.

First, your `test(renderer)` commit carried `Co-Authored-By` and `Claude-Session` trailers. This repo keeps commit messages to the change itself, with no tooling or session metadata in the body or footers, so I amended the message to drop those two lines. The commit is otherwise untouched and still authored by you. Worth setting up locally so it does not come back on the next one.

Second, the test itself landed in the middle of the doc comment for `read_body_capped_rejects_a_chunked_body_over_the_cap`, which left that test undocumented. Moved it down to sit with the other chunked-body tests.

On top of that I pushed one commit that keeps every behaviour fix in this PR and implements a few of them with less code:

- The bounded body read is now `crw_core::body::read_capped`, replacing five near-identical loops. Two already existed in `crw-crawl::sitemap` and `crw-search::client`, and this branch added three more. The helper hands the stream's own error type back, so `crw-search` keeps mapping through `reqwest_message` and a credentialed backend URL stays out of logs.
- That also fixes a bug: `RobotsTxt::fetch` inferred truncation from `bytes.len() == MAX_ROBOTS_BYTES`, which is true of a complete 500 KiB file too, so the trim-to-last-newline dropped a valid final rule from any robots.txt that landed exactly on the cap. Truncation is now reported only when the stream still had bytes to give.
- `RobotsTxt::parse` fills an exact bucket and a wildcard bucket in one pass instead of building `Vec<Group>` and filtering it afterwards. Product-token precedence, agent runs and the empty-exact-group case all behave as before, and all sixteen robots tests pass unchanged.
- The robots fetch in `run_crawl` and `discover_urls` matches in two levels rather than four.

`cargo test --workspace` is 4320 passing, clippy is clean under `-D warnings`.

Nice catch on the query-keyed robots rules and the pool self-deadlock in particular. Those were real.

us:
One more commit, and it reverses a decision this branch made.

`RobotsTxt::fetch` failing was treated as terminal: `run_crawl` marked the job Failed without fetching the seed, and `discover_urls` returned `TargetUnreachable` with zero URLs. That hands any origin whose robots.txt returns a 503 the ability to stop a crawl outright, and every page it never asked us to withhold goes unfetched with it. An origin that cannot serve a file has not forbidden anything.

Measured against a local origin answering 503 on `/robots.txt` and 200 on everything else, robots enforcement on, same binary either way:

    before   error: crawl failed: robots.txt unreachable  (0 pages)
    after    Crawl completed: 2 pages

Both surfaces now log which origin went dark and continue with no rules, which is what they did before this branch. Two follow-ons:

- `fetch` returns `CrwResult<Self>`. The `Option` existed to tell a 4xx from a 5xx, and once every caller proceeds either way there is nothing left to tell apart. A 4xx is an empty rule set; a 5xx stays an error so the caller has something to log.
- `discover_urls` gives the robots fetch half the remaining budget rather than all of it. A hanging robots.txt used to be handed the whole deadline, so the seed and the sitemap probes got nothing left and discovery returned zero URLs, independently of the fail-closed arm.

The three tests that pinned the old behaviour are rewritten rather than deleted, so the new contract is held in the same places: `a_robots_txt_we_cannot_read_does_not_fail_the_crawl`, `a_robots_txt_we_cannot_read_does_not_stop_discovery`, `hanging_robots_leaves_budget_for_the_seed`.

Everything else about your robots work stands: the query-aware matching, the agent-run grouping, the exact product-token precedence and the `directive_value` slicing fix are all unchanged and all their tests pass. `cargo test --workspace` is 4320 passing, clippy clean under `-D warnings`.

PR #558, opened 2026-09-17
feat(scrape): cache the fetch so a repeat scrape skips the network

us:
A page fetched once is reused for an hour. That is what `docs/docs/scraping.md`
has described all along, and the engine never had it. `maxAge` and
`storeInCache` were already arriving from the Firecrawl SDK and were being
ignored, and the compat response hardcoded `cacheState: "miss"`.

## What is cached

The FETCH, not the answer. Extraction options are not part of the key, so a
second pass over the same page with a different schema, different formats or
different selectors reuses one fetch and still runs its own extraction. That is
the multi-pass extraction shape, and it is where the time goes: on a live page
the fetch is ~2.3s and the LLM leg is under a second.

Measured locally against real sites:

| | |
| --- | --- |
| same page, first call | 1327 ms |
| same page again | 315 ms, identical `sourceHash` |
| same page, different output shape | 78 ms, different markdown from the same fetch |
| same page, `maxAge: 0` | 580 ms, fetched, as documented |
| a 2 MB PDF, replayed | 40560 characters both times |

## How it is kept correct

The read sits at the fetch site; the write sits after block classification and
after the JS escalation. Writing at the fetch site would have stored the thin
pre-escalation shell, so exactly the slow pages this exists for would re-run the
chrome render on every hit, and it would have stored anti-bot walls.

Never cached at all: caller headers (that is where a cookie or a token lives),
a caller proxy, a stealth override, a screenshot, or `changeTracking`, which
compares against a live fetch by definition.

Never stored: non-2xx, truncated, deadline-exceeded, blocked, or byte-thin
bodies. A thin body is treated as an unrendered shell whatever the request
asked for, because being *allowed* to escalate is not evidence that escalating
happened: it can be skipped for budget, find no chrome tier, or fail outright.

The budget a fetch ran under is recorded with it, so a page a tight deadline had
to settle for is never handed to a caller who could have done better. The
timestamp is the moment the network answered, not the moment of the write. The
proxy is fingerprinted from the full URL so two sessions on one rotating pool
stay apart, and hashed so credentials never reach the key. A hit that then
escalates writes the better page back and reports `cached: false`, because that
body came off the network.

## Surface

- `maxAge` and `storeInCache` on `/v1/scrape` and on the compat surface, with a
  lenient deserializer there so an SDK sending `"3600000"` as a string still
  works rather than turning into a 400
- `cached` on `/v1` responses, a real `cacheState` on the compat surface
- crawl and map are unchanged: they run their own fetch loop and never enter
  `scrape_url`
- the CLI opts out: a one-shot process cannot hit a cache it just built

## Docs

Two sentences in the scraping docs described something the engine does not do.
One said the cache key is the whole request body; it is the fetch, which gives
callers more hits, not fewer. The other promised per-account isolation, which a
process-global cache in an engine that has no account identity cannot keep.
What replaces it is a rule the code enforces and a test covers: a request
carrying anything caller-specific is never cached at all, so a stored page is
only ever a plain anonymous fetch of a public URL.

24 new tests, 15 on the cache itself and 9 integration tests that assert the
number of requests the origin actually received.


us:
Safety gate before merge, run against the built engine on 14 real pages with an empty cache.

Per URL: one call to seed, one to replay, one with `maxAge: 0` for the live truth.

| | |
| --- | --- |
| served from cache on the second call | 14/14 |
| replays that differ from the page they replayed | 0 |
| median speedup | 3.6x (1.5x to 64.8x) |
| pages that changed between two live fetches | 1 (stripe.com/docs, dynamic, not a cache fault) |

The remaining time on a hit is extraction, which the cache does not remove: it
removes the fetch. A 2 MB PDF replays at 862ms against 1523ms for that reason.

Also verified that a replayed PDF still returns its 40560 characters rather than
the empty document a lost `raw_bytes` would produce, and that a second pass with
a different output shape reuses the fetch (78ms against 1327ms) while producing
its own, different markdown.

